PR #26, opened 2015-10-21
Fix to not login from the same machine in CLI

shishirx34:
Also added a fix to not let users delete the access key to the local machine otherwise it leads us to a weird behavior where you are not authorized because the access key is removed but can't login because you still have the connection info persisted on disk with an invalid access key.

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -84,15 +84,15 @@ function accessKeyList(command: cli.IAccessKeyListCommand): Promise<void> {
         });
 }
 
-function removingLocalAccessKey(): Promise<void> {
+function removeLocalAccessKey(): Promise<void> {
      return Promise<void>((resolve, reject, notify): void => {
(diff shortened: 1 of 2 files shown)


cli/script/command-executor.ts
```diff
@@ -261,9 +271,19 @@ function deserializeConnectionInfo(): IConnectionInfo {
...
+function alreadyLoggedIn(): Promise<void> {
+     return Promise<void>((resolve, reject, notify): void => {
+         log("You are already logged in from this machine!");
+     });
+}
+
 export function execute(command: cli.ICommand): Promise<void> {
     connectionInfo = deserializeConnectionInfo();
 
+    if (connectionInfo && command.type === cli.CommandType.login) {
```
shishirx34: Yes, earlier I had error message shown for register too, but I realized that it isn't really needed since you could just run register multiple times if for some reason you want to have different provider configured. Moving this to switch block.

lostintangent:
LGTM!

PR #27, opened 2015-10-22
Cleaning up package.json

lostintangent:
We were missing the license and keyword fields in our package.json, and the author field wasn't consistent with the other repos (e.g. Cordova and React Native plugins)


dtivel:
LGTM

PR #28, opened 2015-10-22
Adding a little more detail to the root readme

lostintangent:
Our root readme is pretty sparse, so this just adds a little context for folks that navigate to it directly, and includes the right links to get started.


shishirx34:
Lgtm!

silhouettes:
LGTM!

PR #33, opened 2015-10-27
fix: links to SDKs repos

theaqua:


lostintangent:
Thanks for catching this issue! We fixed it in a few other places, but obviously missed a spot :)

PR #35, opened 2015-10-28
CLI should be able to remove apps and deployments named ls/list

shishirx34:
Fix for bug : ```code-push app rm ls``` using our CLI (remove an app named "ls"), the command would fail.

Yargs has some kind of command precedence which checks the usage of the command for an argument in chronological order that it was added to the commands list, so when executing 

'code-push app rm ls' - yargs would interpret it as an ls command and list its usage since we had wrong # of arguments for a ls command. Giving ls, list commands the least precedence enables us to execute rm, remove commands before ls and thereby be able to delete those apps or deployments named 'ls' 'list'.

### DIFF
--- a/cli/script/command-parser.ts
+++ b/cli/script/command-parser.ts
@@ -101,10 +101,10 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
         isValidCommandCategory = true;
         yargs.usage(USAGE_PREFIX + " access-key <command>")
             .demand(/*count*/ 2, /*max*/ 2)  // Require exactly two non-option arguments.
-            .command("list", "List the access keys associated with your account", (yargs: yargs.Argv) => accessKeyList("list", yargs))
-            .command("ls", "List the access keys associated with your account", (yargs: yargs.Argv) => accessKeyList("ls", yargs))
             .command("remove", "Remove an existing access key", (yargs: yargs.Argv) => accessKeyRemove("remove", yargs))
             .command("rm", "Remove an existing access key", (yargs: yargs.Argv) => accessKeyRemove("rm", yargs))
+            .command("list", "List the access keys associated with your account", (yargs: yargs.Argv) => accessKeyList("list", yargs))
+            .command("ls", "List the access keys associated with your account", (yargs: yargs.Argv) => accessKeyList("ls", yargs))
             .check((argv: any, aliases: { [aliases: string]: string }): any => isValidCommand);  // Report unrecognized, non-hyphenated command category.
 
         addCommonConfiguration(yargs);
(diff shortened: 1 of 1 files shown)


silhouettes:
I feel a bit hesitant about this approach, because it relies on some quirks about ordering and the number of parameters for different commands. Also, I feel like there is something about this explanation that doesn't add up, because seemingly by this logic we should have the same problem when running 'code-push app rm add'.

I've been looking at the yargs documentation (https://www.npmjs.com/package/yargs) and I think the problem here is we're using yargs' command() to declare these options, and command() doesn't care about ordering (e.g. 'ls -a -l' is the same as 'ls -l -a'). Have you tried playing with the choices() API? It seems like 'rm', 'add', 'ls' etc are choices for 'app', rather than commands in and of themselves.

priyankurs:
Can u please explain me this all

geof90:
@priyankurs - This was a fix to an issue we were facing, where if you were to do ```code-push app rm ls``` using our CLI, the command would fail, becuase it interprets ```app ls``` as the command to execute instead of ```app rm```, due to command precedence like what @shishirx34 suggests.

@shishirx34 Is this fix general? For example, how does the precedence work between ```add``` and ```rm```? For example, does it work if we do we do ```add rm``` and ```rm add```?

shishirx34:
@priyankurs - As @geof90 explained. Look at the explanation below for root cause.

@silhouettes - let me elaborate. 

When we add commands for app we add them in this order(# demand count) - [add(3), list(2), ls(2), remove(3), rename(5), rm(5)]. (lets call this list1)

Now when ever you run a command, yargs first checks the usage of the command by finding which command you are trying to run(add, list, ls...) it checks this ordered list of commands against all the arguments of the command you are trying to execute(code-push app rm ls => [rm, ls] (lets call this list2)) 

Yargs first checks commands in the list1 against list2(this is the bug/or conscious decision with yargs, I feel it should be list2 checked against list1), if it finds any it checks its usage for demand count, if that succeeds it proceeds further otherwise shows help. Since 'ls' is the preceding command(and the first match in this list comparison) in the list1 above, it checks its demand count(which is 2) however the command you passed has count of '3'(app, rm, ls), thereby showing us the usage help, due to count mismatch(that's the only usage check).

    isValidCommand = true;
    yargs.usage(USAGE_PREFIX + " app " + commandName + " [--format <format>]")
        .demand(/*count*/ 2, /*max*/ 2)  // Require exactly two non-option arguments.
        .example("app " + commandName, "Lists apps in tabular format")
        .example("app " + commandName + " --format json", "Lists apps in JSON format")
        .option("format", { default: "table", demand: false, description: "The output format (\"json\" or \"table\")", type: "string" });

Now when you run 'code-push app rm add', you are right, following this logic we should see the add usage, however the command runs successfully. The reason is the demand count of add(coincidentally is 3) same as rm command, once the usage check(which only checks the demand count) is succeeded our command-parser validates the order of commands and executes them, hence it works fine. Same is true of 'rename' command, if you tried to remove app named 'rename' it will give you help for it since its usage checks demand count of 5.

I agree this fix is kind of a workaround, given that this is such a corner case(only having names of apps & deployments as that of the commands), I strongly feel that it is not worth time/effort finding a better way to fix it. I will open up a bug on yargs for this issue. 

@geof90 - The precedence is in the way I explained above. The commands that are added chronologically are the first whose usage is checked. In case of 'rm add' it will be add and in case of 'add rm' as well it will be add, because 'add' was added before 'rm' command.

geof90:
Thanks for explaining! It does sound like a bug with yargs.

And I see, by making rm precede everything else, we will solve the problem. Sure, there may be some names that cannot be added, but at least there will never be anything that is added that cannot be subsequently removed.

LGTM. :+1:

silhouettes:
Okay - I agree that this particular symptom is a corner case, I just wanted to make sure we fully understood the extent of the issue. If you're confident that the root cause of this issue won't create problems elsewhere then it looks good to me! Thanks so much for explaining.

PR #37, opened 2015-10-30
Bumping the SDK version and fixing some package.json values

lostintangent:
I need to bump the version in order to release the update. While doing that, I noticed a few omissions in the SDK's package.json that I wanted to fix.


dtivel:
LG

geof90:
LGTM

PR #38, opened 2015-11-03
Improve Access Key

geof90:
Generate a better access key that has the account ID mixed into it.

### DIFF
--- a/sdk/script/management/account-manager.ts
+++ b/sdk/script/management/account-manager.ts
@@ -929,15 +929,14 @@ export class AccountManager {
     }
     
     private generateAccessKey(): Promise<string> {
-        return this.getAccountInfo()
-            .then(() => {
-                var newAccessKey: string = uuid.v4();
-                // Strip off all the dashes.
-                newAccessKey = newAccessKey.replace(/-/g, "");
-                // Mix accountID into the key (user should be logged in).
-                newAccessKey = newAccessKey + this.accountId;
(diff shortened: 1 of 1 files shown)


silhouettes:
I implemented some of the changes we talked about, is that okay with you? Basically, the intention is to use the guid for uniqueness, and the crypto random string for unpredictability.

When the two are put together and base64-encoded, the key is 40 characters long. I am tempted to just remove the accountId at the end now, because I imagine it might be easy to confuse your keys with each other if they all end with the same characters, and I'm not sure if the accountId at the end achieves anything. What do you think?

sdk/script/management/account-manager.ts
```diff
@@ -73,7 +74,7 @@ export class AccountManager {
                 req = req.send({ identity: "accessKey" })
```
silhouettes: I noticed that the 'if' in the collapsed section above still has 'Login the old way' as a case. I think we can just remove it, seeing as all of the keys will be migrated?

geof90: Keys will be migrated on the server side, but on the client side, people who used the CLI prior to the change will have them persisted the old way. The format we are persisting them is different, and this code handles that.

silhouettes:
Okay sorry for the hassle, thanks for discussing it with me. Is this okay with you?

geof90:
LGTM. Thanks for taking this over! :)

PR #43, opened 2015-11-04
Prevent date truncation

geof90:
Calling toString() on the date objects prevents the dates from being truncated when the table is printed.

@lostintangent


lostintangent:
LGTM!

PR #45, opened 2015-11-04
code-push logout --local

geof90:
Currently, logout always deletes the access key on the server. Calling it with the additional --local flag removes the access key saved locally, but does not delete the access key on the server so that the same key can be used for future logins.

Bumped the CLI version to 1.1.2-beta, and also updated docs.


cli/README.md
```diff
@@ -54,6 +54,12 @@ After creating the new key, you can specify its value using the `--accessKey` fl
 code-push login --accessKey <accessKey>
 ```
 
+After using the key, if you want to log out of your current session, but still want to be able to re-use the same key for future logins, run the following command:
```
dtivel: "Reuse" is not hyphenated.

dtivel:
looks good

cli/README.md
```diff
@@ -54,6 +54,12 @@ After creating the new key, you can specify its value using the `--accessKey` fl
 code-push login --accessKey <accessKey>
 ```
 
+After using the key, if you want to log out of your current session, but still want to be able to reuse the same key for future logins, run the following command:
```
silhouettes: How about:

If you want to log out of your current session without invalidating the key for future sessions, run the following command:

cli/script/command-parser.ts
```diff
@@ -212,6 +212,12 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
     .command("logout", "Log out of the current session", (yargs: yargs.Argv) => {
         isValidCommandCategory = true;
         isValidCommand = true;
+        yargs.usage(USAGE_PREFIX + " logout [--local <true|false>]")
```
silhouettes: Can we not specify <true|false> here? The convention for these flags is that if they exist, they change something, and if they don't, it's the default behavior

cli/script/command-parser.ts
```diff
@@ -212,6 +212,12 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
+            .demand(/*count*/ 1, /*max*/ 1)  // Require exactly one non-option argument.
+            .example("logout", "Log out and also remove the access key used for the current session.")
```
silhouettes: Could we try:
Log out and also invalidate the access key for future sessions

silhouettes: Actually this is fine, forget I said anything!

cli/script/command-parser.ts
```diff
@@ -212,6 +212,12 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
+            .example("logout --local", "Log out but still allow the use of the same access key for future logins.")
```
silhouettes: Can we remove the word 'still' here?

cli/script/command-parser.ts
geof90: Should I also remove it for --mandatory and --verbose in other commands?

silhouettes: Ah I see, you were matching what was there before. I say yes for sure, the convention in CLI's (e.g. git) is definitely that you don't need to specify 'true' or 'false'.

If you don't mind, could you also update the README?

silhouettes:
Nice work, looks good to me! Sorry for being nitpicky.

PR #49, opened 2015-11-06
When running the deployment list command, show the package by default

silhouettes:
And remove the '--verbose'/'-v' options which previously were used to get the package information.


dtivel:
looks good

geof90:
LGTM

PR #52, opened 2015-11-09
Format JSON output of cli commands to be human readable

silhouettes:


cli/script/command-executor.ts
```diff
@@ -582,6 +582,10 @@ function printDeploymentList(command: cli.IDeploymentListCommand, deployments: D
     }
 }
 
+function printJson(object: any) {
```
dtivel: : void

cli/test/cli.ts
```diff
@@ -7,6 +7,11 @@ import * as cli from "../definitions/cli";
 import * as cmdexec from "../script/command-executor";
 import * as os from "os";
 
+function assertJsonDescribesObject(json: string, object: Object) {
```
dtivel: : void

dtivel:
looks good

silhouettes:
Per prior discussion, I've also removed the word 'minimum' used to describe app store versions in the CLI, as we match the latest update by exact app store version. Thanks dtivel and lostintangent!

PR #53, opened 2015-11-10
Add 'deployment view' command to CLI to show release history

silhouettes:
Here is how it currently looks, I am very open to feedback:

image

Thoughts:
1. Are we happy with the syntax of the command?
2. The 'description' field is not mandatory when you call the 'release' command, hence I wouldn't be surprised if it was empty most of the time. Brainstorm: Make it a compulsory argument to 'release'? Hide the 'description' column if it is completely empty? Leave it as-is?
3. For deployments with no releases, I'm not sure if the output is confusing. Should I use a custom string?

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -16,7 +16,7 @@ import * as yazl from "yazl";
 import wordwrap = require("wordwrap");
 
 import * as cli from "../definitions/cli";
-import { AccessKey, AccountManager, App, Deployment, DeploymentKey } from "code-push";
+import { AccessKey, AccountManager, App, Deployment, DeploymentKey, Package } from "code-push";
 import Promise = Q.Promise;
 
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
@@ -285,8 +285,22 @@ function deploymentRename(command: cli.IDeploymentRenameCommand): Promise<void> 
 
 function deploymentView(command: cli.IDeploymentViewCommand): Promise<void> {
     throwForInvalidOutputFormat(command.format);
+    var storedAppId: string;
 
-    return Q(<void>null);
+    return getAppId(command.appName)
+        .then((appId: string): Promise<string> => {
+            throwForInvalidAppId(appId, command.appName);
+            storedAppId = appId;
+
+            return getDeploymentId(appId, command.deploymentName);
+        })
+        .then((deploymentId: string): Promise<any>[] => {
+            throwForInvalidDeploymentId(deploymentId, command.deploymentName, command.appName);
+
+            return [ sdk.getDeploymentKeys(storedAppId, deploymentId), sdk.getPackageHistory(storedAppId, deploymentId) ];
+        }).spread<void>((deploymentKeys: DeploymentKey[], packageHistory: Package[]): void => {
+            printDeployment(command, deploymentKeys[0], packageHistory);
+        });
 }
 
 function deserializeConnectionInfo(): IStandardLoginConnectionInfo|IAccessKeyLoginConnectionInfo {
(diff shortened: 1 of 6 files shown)


lostintangent:
I don't think we should make the description field mandatory when releasing, but maybe we could make it the rightmost column in the release history table so that its absence doesn't look so confusing/odd. I like the following column order (left to right): Label, Upload Time, App Store Version, Mandatory, Description. I think that might actually make the data more readable.

For deployments with no releases, I agree that we should have a custom string and not bother outputting an empty table. The JSON output should be an empty array.

Regarding the syntax, I can't help but wonder whether "history" wouldn't be a better sub command than "view". "code-push deployment history <appName> <deploymentName>" feels like it reads really well.

lostintangent:
Actually, if we renamed the command to "history", I think you could completely remove the intro sentence from output and display the empty table when there aren't any releases. All of the existing list tasks will display an empty table, so it would be consistent, and if the command was already specific to release history, I don't think the intro sentence is neccessary to explain the output.

dtivel:
"history" is more intuitive and consistent with the command's output.

cli/script/command-executor.ts
```diff
@@ -283,6 +283,26 @@ function deploymentRename(command: cli.IDeploymentRenameCommand): Promise<void>
...
+            throwForInvalidAppId(appId, command.appName);
+            storedAppId = appId;
+
+            return getDeploymentId(appId, command.deploymentName);
+        })
+        .then((deploymentId: string): Promise<any>[] => {
+            throwForInvalidDeploymentId(deploymentId, command.deploymentName, command.appName);
+
+            return [ sdk.getDeploymentKeys(storedAppId, deploymentId), sdk.getPackageHistory(storedAppId, deploymentId) ];
+        }).spread<void>((deploymentKeys: DeploymentKey[], packageHistory: Package[]): void => {
```
dtivel: Put ".spread" on next line.

cli/script/command-executor.ts
```diff
@@ -621,6 +647,26 @@ function printAccessKeys(format: string, keys: AccessKey[]): void {
     }
 }
 
+function printDeployment(command: cli.IDeploymentViewCommand, deploymentKey: DeploymentKey, packageHistory: Package[]): void {
+    packageHistory.reverse(); // Reverse chronological order
```
dtivel: Instead of reversing and then forward iterating, just reverse iterate.

cli/script/command-executor.ts
```diff
@@ -283,6 +283,26 @@ function deploymentRename(command: cli.IDeploymentRenameCommand): Promise<void>
...
+        .then((appId: string): Promise<string> => {
+            throwForInvalidAppId(appId, command.appName);
+            storedAppId = appId;
+
+            return getDeploymentId(appId, command.deploymentName);
+        })
+        .then((deploymentId: string): Promise<any>[] => {
+            throwForInvalidDeploymentId(deploymentId, command.deploymentName, command.appName);
+
+            return [ sdk.getDeploymentKeys(storedAppId, deploymentId), sdk.getPackageHistory(storedAppId, deploymentId) ];
```
dtivel: You don't need to fetch or display the deployment key.  "deployment list" already shows deployment keys.

sdk/script/management/account-manager.ts
```diff
@@ -892,6 +892,36 @@ export class AccountManager {
...
     }
 
+    public getPackageHistory(appId: string, deploymentId: string): Promise<Package[]> {
+        return Promise<Package[]>((resolve, reject, notify) => {
+            var requester = (this._authedAgent ? this._authedAgent : request);
+            var req = requester.get(this.serverUrl + "/apps/" + appId + "/deployments/" + deploymentId + "/packageHistory");
+            this.attachCredentials(req, requester);
+
+            req.end((err: any, res: request.Response) => {
+                    if (err) {
```
dtivel: Indentation is off.

cli/test/cli.ts
```diff
@@ -88,6 +88,28 @@ export class SdkStub {
         }]);
     }
 
+    public getPackageHistory(appId: string, deploymentId: string): Promise<codePush.Package[]> {
+        return Q([
+            <codePush.Package>{
+                appVersion: "1.0.0",
+                isMandatory: false,
+                packageHash: "463acc7d06adc9c46233481d87d9e8264b3e9ffe60fe98d721e6974209dc71a0",
+                blobUrl: "https://codepushstaging.blob.core.windows.net/storagev2/416cfQ5Ge",
```
dtivel: Please hide these blob urls in future, just in case

dtivel:
Looks good.

cli/script/command-executor.ts
silhouettes: That's fine for when it's in tabular format and I have to iterate through it again anyway, but I still want to show it in reverse order when I stringify it for the JSON view. I could reverse it in one case and not the other, but this is just more tidy. It's running on the client anyway so the tiny perf hit won't really matter too much.

silhouettes:
It seems like we've reached consensus, and I'm quite happy with the suggestions too! I've gone and implemented the changes. Thanks for the feedback!

PR #55, opened 2015-11-11
Update README to document the 'deployment history' command

silhouettes:
Forgot to do this earlier. Two thoughts:
- Is the section this is in a little strange?
- Does the 'history' command actually belong under 'deployment'?

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -157,7 +157,7 @@ code-push promote MyApp Staging Production
 
 ### Viewing release history
 
-You can view a history of your releases using the following command:
+You can view a history of releases for a specific app deployment (including promotions) using the following command:
 
 ```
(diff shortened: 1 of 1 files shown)


cli/README.md
```diff
@@ -154,3 +154,11 @@ Once you've tested an update to a specific deployment, and you want to promote i
 code-push promote <appName> <sourceDeploymentName> <destDeploymentName>
 code-push promote MyApp Staging Production
 ```
+
+### Viewing release history
+
+You can view a history of your releases using the following command:
```
lostintangent: What about something like... "In order to view the history of releases for a specific app deployment (including promotions), you can use the following command"

lostintangent:
LGTM!

PR #58, opened 2015-11-17
Login message change

lostintangent:
The `login` and `logout` commands currently output multiple log messages, which make it look a little messy. This change consolidates the general "success" message and the add/remove config file success message into a single message, since the two are conceptually the same anyways. I also cleaned up some of the login text that discusses the browser being launched.

Note that the `logout` command will still display two messages, since the removal of the access key also generates a message, but I didn't want to mess with that right now.


dtivel:
Looks good.

Issue #63, opened 2015-11-23
The CLI doesn't indicate why a release failed when specifying a non-semver compliant app version


lostintangent:
The `code-push release` command displays helpful error messages when the specified app and/or deployment aren't valid, but if you specify a non-semver compliant app version label (e.g. `1.0` instead of `1.0.0`), it just silently fails and displays the help output for the command. I'm assuming that means that the issue is actually with the parser, and that the request is never hitting the server.

The command's examples in its help output correctly displays the use of semver version strings, but we should make sure the CLI provides as much guidance as possible. To that end, it would be great if we could catch this scenario and clearly report the issue to the user so they would know how to respond (e.g. change `1.0` to `1.0.0`).

silhouettes:
On it

PR #65, opened 2015-11-24
Cli improvements

silhouettes:
1. Tidying - use .map() instead of [declare array] -> [forEach] -> [push to array]
2. For the 'app list' command, remove the 'id' column and add a 'Deployments' column with a comma-separated list of deployments
3. Clean up the error message when releasing with a non-semver compliant app store version as in #63

Screenshots:
(Very similar to what actually happened when I was testing)
image

('app list' command at the top, not sure if we should bother clarifying the error message in the second command)
image


lostintangent:
Can we alpha sort the deployment names in the app list? Also, will the new deployment column wrap at some point?

silhouettes:
Nice catch on the wrapping. Yeah I was sorting case sensitively, but maybe an insensitive sort is better. Thanks for the feedback!

PR #67, opened 2015-11-26
Adding extra platform detail

lostintangent:
This PR adds a little more platform specific detail for the `package` and `appVersion` parameters of the `release` command, that way, the docs are a little more self-serving, without needing to refer to the Cordova/React Native SDK docs in order to understand them fully.


geof90:
LGTM!

PR #68, opened 2015-11-29
CLI Tweaks

silhouettes:
1. Show package history in chronological order -  when the package history is long, terminals will naturally scroll to the bottom. Hence, we should display the latest packages last for the user's convenience.
2. Do not show Package column for newly created deployments
3. Update README to mention the deployment history limit of 50 releases


lostintangent:
LGTM!

dtivel:
Looks good.

PR #69, opened 2015-11-30
Minor cli readme fix to make the platforms right in the table.

rmevans9:
Just fixing a minor thing I noticed in the README


lostintangent:
Good catch @rmevans9! Thanks for this contribution. I'll merge this now and will update the docs on the website with this fix as well.

PR #71, opened 2015-12-01
Fix unzipping errors

silhouettes:
We have been logging server errors in our unzip module, and it is highly likely that the cause is users who are uploading their own .zip files. The CLI will zip up a folder for you if you specify the path, but it will leave single files unchanged.

To address this:
1. Modify the CLI to reject .zip uploads with a helpful error message explaining that the user should specify the directory directly (feedback on the error message welcome).
2. Currently it would be impossible to track whether this fixed the issue or not as older CLI versions will still generate these errors. So, send the CLI version in the User-Agent header with every request so that we can track which CLI versions are generating the errors on the server side (also this is just generally useful for telemetry).

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -19,9 +19,11 @@ import wordwrap = require("wordwrap");
 
 import * as cli from "../definitions/cli";
 import { AccessKey, AccountManager, App, Deployment, DeploymentKey, Package } from "code-push";
+var packageJson = require("../package.json");
 import Promise = Q.Promise;
 
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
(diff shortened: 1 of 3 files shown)


lostintangent:
LGTM!

cli/script/command-executor.ts
```diff
@@ -734,7 +736,9 @@ function promote(command: cli.IPromoteCommand): Promise<void> {
 }
 
 function release(command: cli.IReleaseCommand): Promise<void> {
-    if (semver.valid(command.appStoreVersion) === null) {
+    if (command.package.search(/\.zip$/) !== -1) {
```
dtivel: This should be case-insensitive.

silhouettes: Good catch, pushed the fix to master.

PR #76, opened 2015-12-03
CLI rollback feature

silhouettes:
This feature allows you to undo your most recent release if you have shipped a broken or unintended update. It is equivalent to performing a new release with the same package contents and metadata as the release preceding your most recent one.

For performance and simplicity reasons this feature is implemented on the server side, with the implementation here being a simple POST request. In future, there is quite a lot of duplication and boilerplate in this project that can probably be refactored out (not the highest priority right now).

### DIFF
--- a/cli/definitions/cli.ts
+++ b/cli/definitions/cli.ts
@@ -14,9 +14,9 @@
     login,
     logout,
     promote,
-    rollback,
     register,
-    release
+    release,
+    rollback
 }
 
 export interface ICommand {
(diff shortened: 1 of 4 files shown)


cli/definitions/cli.ts
```diff
@@ -14,6 +14,7 @@
     login,
     logout,
     promote,
+    rollback,
```
dtivel: Sort

cli/script/command-executor.ts
```diff
@@ -401,6 +401,9 @@ export function execute(command: cli.ICommand): Promise<void> {
                 case cli.CommandType.promote:
                     return promote(<cli.IPromoteCommand>command);
 
+                case cli.CommandType.rollback:
+                    return rollback(<cli.IRollbackCommand>command);
```
dtivel: Sort

dtivel:
Looks good.

lostintangent:
LGTM!

silhouettes:
Feedback:
- Specify which label was rolled back from
- Display a confirmation dialog

shishirx34:
LGTM!

PR #77, opened 2015-12-03
Fix to assign packageJson to pkg.

shishirx34:
I tested with pkg passed to the updateNotifier as options object and renamed it to packageJson which broke the behavior. 

Also removed extra require to use packageJson in version for ascii art.


silhouettes:
Looks good!

PR #79, opened 2015-12-05
Add documentation on rollback command

silhouettes:
Also update screenshots.


cli/README.md
```diff
@@ -184,12 +184,28 @@ code-push promote <appName> <sourceDeploymentName> <destDeploymentName>
...
 ```
 
+The release produced by a promotion will be annotated in the output of the `deployment history` command.
+
+## Rolling back undesired updates
+
+If you release an update that is broken or contains unintended features, it is easy to roll it back using the `rollback` command:
+
+```
+code-push rollback MyApp Production
```
lostintangent: It seems obvious, but could we add a "formal" example of the args here like we do for the promote command? (E.g. "code-push rollback <appName> <deploymentName>")

lostintangent:
LGTM!

Issue #83, opened 2015-12-13
Rollback command should allow specifying a target release label


lostintangent:
Currently, the rollback command only supports rolling back to the previous release. While this accommodates the most common use case, it's possible that a user tries to solve a bad release with another release, which also ends up having issues, only to realize that they wanted to just rollback to the release before the initial bad one. Currently that wouldn't be possible, and therefore, the dev would have to try to fix the release and update a new one again, which is an extremely error prone operation, especially during a fire drill situation

Since the goal of the rollback command is to provide a simple and reliable way to address fire drills associated with a bad release, we should accommodate rolling back to an arbitrary release. Even if that only accommodates a small percentage of scenarios, it would be worth it, so that devs aren't left hanging after trying to fix an issue.

We should add a new optional flag to the rollback command called "--targetRelease" (which is aliased as "-r"), that takes the release label (e.g. v23) that the user wants to rollback to. When this label isn't specified, the rollback command would behave the way it already does. Otherwise, the command should create a new release whose contents and metadata represent the target release.

If the target release flag is specified, and the specified release label is invalid or doesn't exist within the deployment, a clear error should be displayed in the CLI.

PR #85, opened 2015-12-17
Allow rollbacks to an arbitrary release label

silhouettes:
Adding support as per #83. Relies on server-side support which has been implemented!

### DIFF
--- a/cli/script/command-parser.ts
+++ b/cli/script/command-parser.ts
@@ -189,7 +189,7 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
             .demand(/*count*/ 3, /*max*/ 3)  // Require exactly three non-option arguments.
             .example("rollback MyApp Production", "Perform a rollback on the \"Production\" deployment of \"MyApp\"")
             .example("rollback MyApp Production --targetRelease v4", "Perform a rollback on the \"Production\" deployment of \"MyApp\" to the v4 release.")
-            .option("targetRelease", { alias: "r", default: null, demand: false, description: "The label of the release to be rolled back to", type: "string" });
+            .option("targetRelease", { alias: "r", default: null, demand: false, description: "The label of the release to be rolled back to (e.g. v4)", type: "string" });
 
         addCommonConfiguration(yargs);
     })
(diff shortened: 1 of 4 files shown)


cli/script/command-parser.ts
```diff
@@ -185,10 +185,12 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
     })
     .command("rollback", "Performs a rollback on the latest package of a specific deployment", (yargs: yargs.Argv) => {
-        yargs.usage(USAGE_PREFIX + " rollback <appName> <deploymentName>")
+        yargs.usage(USAGE_PREFIX + " rollback <appName> <deploymentName> [--targetRelease <releaseLabel>]")
             .demand(/*count*/ 3, /*max*/ 3)  // Require exactly three non-option arguments.
-            .example("rollback MyApp Production", "Perform a rollback on the \"Production\" deployment of \"MyApp\"");
-            
+            .example("rollback MyApp Production", "Perform a rollback on the \"Production\" deployment of \"MyApp\"")
+            .example("rollback MyApp Production --targetRelease v4", "Perform a rollback on the \"Production\" deployment of \"MyApp\" to the v4 release.")
+            .option("targetRelease", { alias: "r", default: null, demand: false, description: "The label of the release to be rolled back to", type: "string" });
```
shishirx34: do you think it is intuitive to find the release label? or may be put in the command to list the release labels in the description here?

sdk/script/management/account-manager.ts
```diff
@@ -749,10 +749,10 @@ export class AccountManager {
         });
     }
 
-    public rollbackPackage(appId: string, deploymentId: string): Promise<void> {
+    public rollbackPackage(appId: string, deploymentId: string, targetRelease?: string): Promise<void> {
         return Promise<void>((resolve, reject, notify) => {
             var requester = (this._authedAgent ? this._authedAgent : request);
-            var req = requester.post(this.serverUrl + "/apps/" + appId + "/deployments/" + deploymentId + "/rollback");
+            var req = requester.post(this.serverUrl + "/apps/" + appId + "/deployments/" + deploymentId + "/rollback/" + (targetRelease || ""));
```
shishirx34: I prefer having this as a query string since it is optional.

shishirx34:
LGTM!

cli/script/command-parser.ts
lostintangent: I think of this as an edge case scenario, so I think it's fine to not overly document how to get the label. That said, maybe the description can have an "(e.g. v12)" example to make it a little clearer.

lostintangent:
LGTM!

PR #86, opened 2015-12-22
Attach label to updateCheck requests

geof90:
This allows the server to use the label to figure out the current version of the package instead of searching the history for the packageHash.


shishirx34:
LGTM

silhouettes:
Looks good!

PR #91, opened 2015-12-29
Remove package hashes in "deployment list" command

geof90:
You can still see the hashes using the "--format json" flag, if needed.


lostintangent:
LGTM!

PR #92, opened 2015-12-30
Doc updates

lostintangent:
This PR improves our docs in the following ways:

1. Documents the new `--targetRelease` parameter of the `rollback` command.
2. Documents the ability to include newlines in the `--description` parameter of the `release` command.
2. Includes more description about how the `mandatory` flag of a release is handled by the server.
3. Includes more description about how the `promote` and `rollback` commands work.
4. Includes more prescription about creating separate CodePush apps when targeting both Android and iOS
5. Includes more description about how to think about deployments.
6. Includes more prescription about how to use `Staging` for testing updates before releasing to `Production.
7. Includes an example of how to prepare/release a React Native iOS app that uses assets

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -79,16 +79,16 @@ Before you can deploy any updates, you need to register an app with the CodePush
 code-push app add <appName>
 ```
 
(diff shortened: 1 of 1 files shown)


geof90:
LGTM!

PR #93, opened 2015-12-30
MOD: In the CodePush Documentation, the description is optional, but …

shawnXiao:
…the backslash module must have str parameter


geof90:
Looks good! Thanks @shawnXiao!

lostintangent:
Looks good. We should release a bug fix update for this.

PR #95, opened 2015-12-30
Adding detail for store version

lostintangent:
This PR adds a little more detail about the app store version parameter, and makes a few other minor doc tweaks.


geof90:
LGTM!

PR #97, opened 2016-01-06
Update REST definitions

dtivel:
Copy updated REST definitions from CodePush service.
Fix up test code after making some properties optional.


silhouettes:
Looks good!

PR #100, opened 2016-01-12
Authentication v2

silhouettes:
(Merging into 'breaking' branch)

A faster, more secure, and simpler login process using the access-key as a bearer token with every request instead of exchanging it for a session identifier cookie. Also, a significant amount of simplifying refactoring work, especially to the SDK for future external consumption.

Additional SDK changes:

1. Remove the concepts of 'logging in' and 'logging out' as the server now maintains no session state. The SDK is initialized with an access-key and that is sent with every request. Simply let the SDK object be garbage collected if you are done (and optionally delete the access-key on the server).
2. Remove the 'ConnectionInfo' type and the concept of a base 64-encoded 'Access Token' from the SDK. The SDK should only care about the access-key, with the other stuff being CLI-specific.
3. Use the 'Accept' header to indicate the desired server API version.
4. Tidy up exposed public members and constructor interface.

Note that the external OAuth provider login/registration process is not currently covered in the SDK - the CLI has always independently launched an external browser session. I may do future investigative work to see if there are simplifications we can make here for external consumption.


sdk/script/management/account-manager.ts
```diff
@@ -857,27 +736,19 @@ export class AccountManager {
...
-        } else {
-            request.withCredentials();
-        }
-
-        if (this._userAgent) {
-            request.set("User-Agent", this._userAgent);
-        }
+    private attachCredentials(request: superagent.Request<any>): void {
+        request.set("User-Agent", this._userAgent);
+        request.set("Accept", "application/vnd.code-push." + AccountManager.API_VERSION + "+json");
```
lostintangent: Doesn't TypeScript support string templates now?

sdk/script/management/account-manager.ts
```diff
@@ -410,14 +313,13 @@ export class AccountManager {
...
             var app = <App>{ name: appName };
-            var requester = (this._authedAgent ? this._authedAgent : request);
 
-            var req = requester.post(this.serverUrl + "/apps/");
-            this.attachCredentials(req, requester);
+            var request: superagent.Request<any> = superagent.post(this._serverUrl + "/apps/");
+            this.attachCredentials(request);
 
-            req.set("Content-Type", "application/json;charset=UTF-8")
+            request.set("Content-Type", "application/json;charset=UTF-8")
```
lostintangent: Are we using this content type enough to justify making it a constant somewhere?

sdk/script/management/account-manager.ts
```diff
@@ -32,123 +34,37 @@ interface PackageToUpload {
     isMandatory: boolean;
 }
 
-interface ILoginInfo {
-    accessKeyName: string;
-    providerName: string;
-    providerUniqueId: string;
-}
-
 export class AccountManager {
```
lostintangent: What do you think about making this the default export?

lostintangent:
I deleted my two other comments since they're not necessary, but it would be cool if we made the `AccountManager` the default export. Other than that, this LGTM! Awesome job dude. This is so much cleaner than before :D

cli/script/command-executor.ts
```diff
@@ -532,52 +515,56 @@ function loginWithAccessTokenInternal(serverUrl: string): Promise<void> {
                 return;
             }
 
-            if (!accessToken) {
+            var connectionInfo: ILegacyLoginConnectionInfo|ILoginConnectionInfo;
+            try {
+                var decoded: string = base64.decode(accessToken);
+                connectionInfo = JSON.parse(decoded);
+            } catch (ex) {
+            }
```
dtivel: With existing helper methods, you can simplify lines 518-523 to:

    var decoded: string = tryBase64Decode(accessToken);
    var connectionInfo: ILegacyLoginConnectionInfo|ILoginConnectionInfo = tryJSON(decoded);

cli/script/command-executor.ts
```diff
@@ -334,35 +303,45 @@ function deserializeConnectionInfo(): IStandardLoginConnectionInfo|IAccessKeyLog
...
-    return loginWithAccessToken()
-        .then((): Promise<void> => {
             switch (command.type) {
+                case cli.CommandType.login:
+                    return login(<cli.ILoginCommand>command);
+
+                case cli.CommandType.register:
+                    return register(<cli.IRegisterCommand>command);
+
+                case cli.CommandType.logout:
```
dtivel: Sort these.

cli/script/command-executor.ts
```diff
@@ -882,7 +869,7 @@ function requestAccessToken(): Promise<string> {
 function serializeConnectionInfo(serverUrl: string, accessToken: string): void {
     // The access token should have been validated already (i.e.:  logging in).
     var json: string = tryBase64Decode(accessToken);
-    var standardLoginConnectionInfo: IStandardLoginConnectionInfo = tryJSON(json);
+    var standardLoginConnectionInfo: ILegacyLoginConnectionInfo = tryJSON(json);
```
dtivel: legacyLoginConnectionInfo

cli/script/command-executor.ts
```diff
@@ -882,7 +869,7 @@ function requestAccessToken(): Promise<string> {
...
 
     if (standardLoginConnectionInfo) {
         // This is a normal login.
```
dtivel: Update comment.

cli/script/command-executor.ts
```diff
@@ -891,7 +878,7 @@ function serializeConnectionInfo(serverUrl: string, accessToken: string): void {
         fs.writeFileSync(configFilePath, json, { encoding: "utf8" });
     } else {
         // This login uses an access token
```
dtivel: access key

cli/script/command-executor.ts
```diff
@@ -882,7 +869,7 @@ function requestAccessToken(): Promise<string> {
 function serializeConnectionInfo(serverUrl: string, accessToken: string): void {
```
dtivel: accessTokenOrKey

sdk/script/management/account-manager.ts
```diff
@@ -857,27 +736,19 @@ export class AccountManager {
...
-        }
+    private attachCredentials(request: superagent.Request<any>): void {
+        request.set("User-Agent", this._userAgent);
+        request.set("Accept", "application/vnd.code-push." + AccountManager.API_VERSION + "+json");
+        request.set("Authorization", "Bearer " + this._accessKey);
     }
 
     private generateAccessKey(): Promise<string> {
-        return this.getAccountInfo().then(() => {
+        return this.getAccountInfo().then((account: Account) => {
```
dtivel: Put continuation on new line.

cli/script/command-executor.ts
```diff
@@ -334,35 +303,45 @@ function deserializeConnectionInfo(): IStandardLoginConnectionInfo|IAccessKeyLog
...
-
-function notifyAlreadyLoggedIn(): Promise<void> {
-    return Q.fcall(() => { throw new Error("You are already logged in from this machine."); });
+    var connectionInfo: ILegacyLoginConnectionInfo|ILoginConnectionInfo = tryJSON(savedConnection);
+    return connectionInfo;
 }
 
 export function execute(command: cli.ICommand): Promise<void> {
-    connectionInfo = deserializeConnectionInfo();
+    var connectionInfo = deserializeConnectionInfo();
```
dtivel: : ILegacyLoginConnectionInfo|ILoginConnectionInfo

sdk/test/management-sdk.ts
```diff
@@ -11,7 +11,7 @@ var manager: AccountManager;
 describe("Management SDK", () => {
 
     beforeEach(() => {
-        manager = new AccountManager(/*serverUrl=*/ "http://localhost", /*userAgent=*/ "unit-test/1.0.0");
+        manager = new AccountManager(/*accessKey=*/ "stub", /*userAgent=*/ "unit-test/1.0.0", /*serverUrl=*/ "http://localhost");
```
dtivel: Perhaps some form of "access key" instead of "stub."  A stub returns a canned response.  The response itself is not a stub, so having AccountManager.accessKey return "stub" is odd.

cli/script/command-executor.ts
```diff
@@ -511,11 +490,15 @@ function initiateExternalAuthenticationAsync(serverUrl: string, action: string):
...
-        sdk = new AccountManager(command.serverUrl, userAgent);
-        return sdk.loginWithAccessToken(command.accessKey)
-            .then((): void => {
-                // The access token is valid.
-                serializeConnectionInfo(command.serverUrl, command.accessKey);
+        sdk = new AccountManager(command.accessKey, userAgent, command.serverUrl);
+        return sdk.isAuthenticated()
+            .then((isAuthenticated: boolean): void => {
+                if (isAuthenticated) {
+                    // The access token is valid.
```
dtivel: The comment is now superfluous.

cli/script/command-executor.ts
```diff
@@ -891,7 +878,7 @@ function serializeConnectionInfo(serverUrl: string, accessToken: string): void {
         fs.writeFileSync(configFilePath, json, { encoding: "utf8" });
     } else {
         // This login uses an access token
-        var accessKeyLoginConnectionInfo: IAccessKeyLoginConnectionInfo = { serverUrl: serverUrl, accessKey: accessToken };
+        var accessKeyLoginConnectionInfo: ILoginConnectionInfo = { serverUrl: serverUrl, accessKey: accessToken };
         json = JSON.stringify(accessKeyLoginConnectionInfo);
         fs.writeFileSync(configFilePath, json, { encoding: "utf8" });
```
dtivel: Lines 878 and 883 can be deduplicated by moving after the `if...else` block.  Might also be able to deduplicate the JSON.stringify(...) call like so:

    var connectionInfo: IStandardLoginConnectionInfo | IAccessKeyLoginConnectionInfo = tryJSON(json);

    if (connectionInfo) {
        connectionInfo.serverUrl = serverUrl;
    } else {
        connectionInfo = { serverUrl: serverUrl, accessKey: accessToken };
    }

    json = JSON.stringify(connectionInfo);
    fs.writeFileSync(configFilePath, json, { encoding: "utf8" });

cli/script/command-executor.ts
```diff
@@ -26,14 +26,14 @@ var progress = require("progress");
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
 var userAgent: string = packageJson.name + "/" + packageJson.version;
```
dtivel: You could remove this since AccountManager sets this value by default.

cli/script/command-executor.ts
```diff
@@ -115,13 +88,9 @@ function accessKeyList(command: cli.IAccessKeyListCommand): Promise<void> {
...
 }
 
-function removeLocalAccessKey(): Promise<void> {
-    return Q.fcall(() => { throw new Error("Cannot remove the access key for the current session. Please run 'code-push logout' if you would like to remove this access key."); });
-}
-
 function accessKeyRemove(command: cli.IAccessKeyRemoveCommand): Promise<void> {
-    if (connectionInfo && (command.accessKeyName === (<IStandardLoginConnectionInfo>connectionInfo).accessKeyName || command.accessKeyName === (<IAccessKeyLoginConnectionInfo>connectionInfo).accessKey)) {
-        return removeLocalAccessKey();
+    if (command.accessKeyName === sdk.accessKey) {
```
dtivel: It's worth renaming accessKeyName to accessKey.

dtivel:
Looks good.

shishirx34:
Looks good.

cli/script/command-executor.ts
silhouettes: This sets the user-agent to be `code-push-cli/<CLI-version>`, whereas the default SDK user agent is `code-push/<SDK-version>`. I think it's useful to distinguish the CLI as a distinct user agent, especially when the SDK will be consumed externally. But, I can also see the simplicity benefit of having user-agent strings being defined *only* by the SDK version, so feel free to discuss if you think that should be the case!

silhouettes:
The default export thing is actually a surprisingly hard change to make considering that we need to handle exporting both TypeScript interfaces and a default JavaScript object, manage .d.ts generation, and handle inter-op between CommonJS and ES6. I've talked to Jonathan about this and I'll be doing this as part of a later PR to prepare the SDK for external usage.

PR #101, opened 2016-01-13
Report Acquisition Status

geof90:
Goal: To allow client devices to better report the status of their update acquisition (whether an update is successfully installed or rolled back, which update is the client running). The eventual goal is to allow developers to see this data from the CLI so that they can better manage their updates!

This PR: Mainly includes changes to the acquisition SDK
   - The config passed to the constructor for the AcquisitionManager must now include the native app version and a unique ID for the client (e.g. device GUID).
   - The reportStatus() method takes now two optional parameters - the current installed package, and the status. If both are not specified, it will be reported to the server that the app is running the version shipped with the app store binary. If both are specified, then depending on the status, it will be reported to the server that the app has either installed a new version, or that the current version has failed to install (rolled back due to an error).

### DIFF
--- a/sdk/script/acquisition-sdk.ts
+++ b/sdk/script/acquisition-sdk.ts
@@ -46,9 +46,19 @@ export interface LocalPackage extends Package {
 export interface Callback<T> { (error: Error, parameter: T): void; }
 
 export interface Configuration {
-    serverUrl: string;
+    appVersion: string;
+    clientUniqueID: string;
     deploymentKey: string;
+    serverUrl: string;
     ignoreAppVersion?: boolean
+}
+
+interface DeploymentStatusReport {
+    appVersion: string;
+    clientUniqueID: string;
+    deploymentKey: string; 
+    label?: string;
+    status?: string
 }
 
 export class AcquisitionStatus {
@@ -57,6 +67,8 @@ export class AcquisitionStatus {
 }
 
 export class AcquisitionManager {
+    private _appVersion: string;
+    private _clientUniqueID: string;
     private _deploymentKey: string;
     private _httpRequester: Http.Requester;
     private _ignoreAppVersion: boolean;
@@ -70,6 +82,8 @@ export class AcquisitionManager {
             this._serverUrl += "/";
         }
 
+        this._appVersion = configuration.appVersion;
+        this._clientUniqueID = configuration.clientUniqueID;
         this._deploymentKey = configuration.deploymentKey;
         this._ignoreAppVersion = configuration.ignoreAppVersion;
     }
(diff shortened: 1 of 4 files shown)


geof90:
cc @dlebu

sdk/script/acquisition-sdk.ts
```diff
@@ -134,25 +148,37 @@ export class AcquisitionManager {
...
-                url += "/deploy";
-                body = JSON.stringify({ deploymentKey: this._deploymentKey, status: status, message: message });
-                break;
-
-            default:
-                if (callback) {
-                    callback(new Error("Unrecognized status."), /*not used*/ null);
-                }
-                return;
+    public reportStatus(package?: Package, status?: string, callback?: Callback<void>): void {
```
lostintangent: Should we have constants for the status values or make it an enum? Currently, it seems like the consumer of this API needs to know what value represents success and failure, and it might make sense to have the acquisition SDK do this.

sdk/script/acquisition-sdk.ts
```diff
@@ -46,17 +46,29 @@ export interface LocalPackage extends Package {
...
 export interface Configuration {
-    serverUrl: string;
+    appVersion: string;
+    clientUniqueID: string;
     deploymentKey: string;
+    serverUrl: string;
     ignoreAppVersion?: boolean
 }
 
+interface DeploymentStatusReport {
```
silhouettes: Is this interface synchronized with the server? Would it be worth adding this to rest-definitions?

sdk/script/acquisition-sdk.ts
silhouettes: He has AcquisitionStatus exported above, which is equivalent to an enum (minus type-checking I guess, but at least for the REST API interface I feel like string constants are a better choice than TypeScript enums, which are integer constants).

We could make this function private, and then introduce public reportSuccess() and reportFailure() functions that call this internally?

silhouettes:
Looks good!

sdk/script/acquisition-sdk.ts
```diff
@@ -134,25 +148,37 @@ export class AcquisitionManager {
...
+                case AcquisitionStatus.DeploymentFailed:
+                    body.status = status;
+                    break;
+
+                default:
+                    if (callback) {
+                        if (!status) {
+                            callback(new Error("Missing status argument."), /*not used*/ null);
+                        } else {
+                            callback(new Error("Unrecognized status" + status + "."), /*not used*/ null);
```
dtivel: `...status<space>" + status ...`

sdk/script/acquisition-sdk.ts
```diff
@@ -46,17 +46,29 @@ export interface LocalPackage extends Package {
 export interface Callback<T> { (error: Error, parameter: T): void; }
 
 export interface Configuration {
-    serverUrl: string;
+    appVersion: string;
+    clientUniqueID: string;
```
dtivel: Globally clientUniqueId not clientUniqueID.

sdk/script/acquisition-sdk.ts
dtivel: Originally this was intended to be a general purpose reporting API.  In light of these changes rename it to reportDeploymentStatus.

sdk/script/acquisition-sdk.ts
dtivel: Yes, please add to rest-definitions.d.ts in both the service repo and here.

sdk/script/acquisition-sdk.ts
dlebu: Renaming the function could break the clients. We should keep the changes to a minimum.

dlebu: Given that the arguments are changed, it is probably a good thing to rename it, and to leave the old method in place for a while for compatibility reasons.

dtivel:
Looks good.

dlebu:
You still need to bump the versions in plugin.xml and package.json.

sdk/script/acquisition-sdk.ts
silhouettes: I might be wrong, but as long as the semver version is bumped appropriately I think he can change these methods how he pleases, because the SDK version referenced by the client plugins is fixed.

geof90:
@dlebu We're gonna wait for some other changes e.g the app collaboration feature as well as other management SDK changes to support viewing metrics before bumping the version.

dlebu:
@geof90 @dtivel I just confirmed that the Cordova CLI installs the latest version of the sdk, regardless of the beta status. However, I did a little test with the renamed method and the reporting fails gracefully - the exception is caught and ignored. Thus, there is no need for keeping the old function around. @lostintangent

dlebu:
LGTM!

PR #106, opened 2016-01-18
reportStatusDownload

geof90:
Small change. In the cordova/RN SDKs, if the install mode is ON_NEXT_RESTART, there could be a large gap of time between the time a package is downloaded and the time a package is finally installed. This adds a new acquisition API, "reportStatusDownload" to be used to report to the server whenever the client downloads a package, and will be called the moment a package is downloaded.

### DIFF
--- a/definitions/rest-definitions.d.ts
+++ b/definitions/rest-definitions.d.ts
@@ -15,6 +15,11 @@ declare module "rest-definitions" {
         status?: string
     }
 
+    export interface DownloadReport {
+        deploymentKey: string; 
+        label: string;
+    }
+
     export interface PackageInfo {
         appVersion: string;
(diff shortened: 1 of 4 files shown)


lostintangent:
LGTM! What do you think about sending the device ID along with the download report so that we have the option to use it in the future without needing to rev the client SDKs? I can't think of a reason to need it now, so I wouldn't block on it, but I'd love to have options for future improvements on the server in the future.

dlebu:
I was not aware of this change. Please add me to the PR on all future SDK changes.

dlebu:
What about the case when download failed? Wouldn't it make sense to have a status member for the `DownloadReport` interface?

PR #107, opened 2016-01-19
Display acquisition metrics

geof90:
This changes allows developers to see metrics for the acquisition of CodePush updates via the "deployment list/ls", "deployment history/h" and a new "deployment metrics" command.

screen shot 2016-01-19 at 11 38 13 pm


cli/definitions/cli.ts
```diff
@@ -11,6 +11,7 @@
     deploymentRemove,
     deploymentRename,
     deploymentHistory,
+    deploymentMetrics,
```
dtivel: Sort all.

cli/script/command-executor.ts
```diff
@@ -398,6 +471,9 @@ export function execute(command: cli.ICommand): Promise<void> {
 
                 case cli.CommandType.deploymentHistory:
                     return deploymentHistory(<cli.IDeploymentHistoryCommand>command);
+                    
+                case cli.CommandType.deploymentMetrics:
```
dtivel: Sort all cases.

cli/script/command-executor.ts
```diff
@@ -498,6 +574,10 @@ function getDeploymentId(appId: string, deploymentName: string): Promise<string>
         });
 }
 
+function getLabelMetricKey(label: string, metricType: string) {
```
dtivel: : string

cli/script/command-executor.ts
```diff
@@ -619,12 +699,14 @@ function printDeploymentList(command: cli.IDeploymentListCommand, deployments: D
...
         if (showPackage) {
             headers.push("Package Metadata");
+            headers.push("Metrics");
         }
         printTable(headers, (dataSource: any[]): void => {
             deployments.forEach((deployment: Deployment, index: number): void => {
                 var row = [deployment.name, deploymentKeys[index]];
                 if (showPackage) {
                     row.push(getPackageString(deployment.package));
+                    row.push(getPackageMetricsString(<PackageWithMetrics>(deployment.package)))
```
dtivel: ;

cli/script/command-executor.ts
```diff
@@ -657,7 +739,69 @@ function printDeploymentHistory(command: cli.IDeploymentHistoryCommand, packageH
...
+                ]);
+            });
+        });
+    }
+}
+
+function printDeploymentMetrics(command: cli.IDeploymentMetricsCommand, metrics: any): void {
+    var labelToMetricsMap: { [label: string] : DeploymentMetric } = {};
+    var totalActive: number = 0;
+    var versionRegex = /v(\d+)/;
```
dtivel: `/^v\d+$/`

cli/test/cli.ts
```diff
@@ -387,12 +401,18 @@ describe("CLI", () => {
...
                             appVersion: "1.0.0",
                             description: "fgh",
-                            label: "ghi",
+                            label: "v2",
                             packageHash: "jkl",
                             isMandatory: true,
                             size: 10,
                             blobUrl: "http://mno.pqr",
-                            uploadTime: +1000
+                            uploadTime: +1000,
```
dtivel: Remove +

cli/script/command-executor.ts
```diff
@@ -657,7 +739,69 @@ function printDeploymentHistory(command: cli.IDeploymentHistoryCommand, packageH
...
+                break;
+        }
+    });
+    
+    if (command.format === "json") {
+        printJson(labelToMetricsMap);
+    } else if (command.format === "table") {
+        printTable(["Label", "Active", "Downloaded", "Installed", "Failed"], (dataSource: any[]) => {
+            Object.keys(labelToMetricsMap).forEach((label: string) => {
+                var isCodePushLabel: boolean = versionRegex.test(label);
```
dtivel: isLabel seems sufficient.

sdk/script/management/account-manager.ts
```diff
@@ -596,6 +596,36 @@ export class AccountManager {
         });
     }
 
+    public getDeploymentMetrics(appId: string, deploymentId: string) {
```
dtivel: `: Promise<any>`

Also fix getDeployment(...) above (line 569).

dtivel:
Looks good.

silhouettes:
I have a few thoughts on the presentation of the data, other thoughts welcome:

1. In the `deployment metrics` command, the label `1.2.3` is a little bit unclear - would it be worth annotating it somehow, e.g. `1.2.3 (app store version)` or `1.2.3 (base)` or adding an appVersion column and just making it `base`?
2. In the `deployment history` command, I think percentages are more useful than raw numbers, especially if we can get the raw data with the `deployment metrics` command. I think instead of raw success/failure numbers, it would be good to have a single success percentage, seeing as we have the raw download number there.
3. Similarly in `deploment history`, I think instead of a raw 'Active' number it should be a percentage, and it should be listed in brackets next to the label, e.g. `v1 (0.1%)`. This may be more controversial though.
4. I think bad releases should pop out, one possible way is that if the failure rate is above a certain percentage, it will be colored red.

lostintangent:
Discussing Richard's comments above:

1. I agree. I'm in favor of adding a "App store version" column to the metrics table, just like we have in the history table. It's really the binary version + label that uniquely identifies the code-base, so I think it's relevant here. Additionally, this would allow us to have the "Label" column be blank as a potentially indicator that the row indicates a binary version and not a CodePush update.

2. Fully agreed. @geof90 and I talked about this already. Percentages are king here.

3. Hmm, not sure I'm in love with putting the active number by the lable.

4. Agreed

5. I'd actually be in favor of not showing the failure key/pair at all unless it's non-zero. And when we show it, we should it in red.

The following is how I've roughly envisioned displaying this data:

Active: 34.34% (12,344)
Installs: 3,345 (56 pending)
Failures: 23

There are primarily three questions that I'd like devs to be able to answer very easily/quickly:

1. "I accidentally released an update that I didn't mean to. Has anyone actually downloaded it yet? If not, I can confidently do a rollback and no my mistake was mitigated." Showing 0 installs and 0 downloads can help answer this.

2. "Have any of my end-users attempted to install the update and it failed?" The failed count is obviously the key point here. We should make this extremely visible when it's non-zero.

3. "How many of my users have upgraded to the latest?" The active percentage and count helps with this.

cli/script/command-executor.ts
```diff
@@ -317,11 +357,44 @@ function deploymentHistory(command: cli.IDeploymentHistoryCommand): Promise<void
...
         .then((deploymentId: string): Promise<Package[]> => {
             throwForInvalidDeploymentId(deploymentId, command.deploymentName, command.appName);
+            storedDeploymentId = deploymentId;
 
             return sdk.getPackageHistory(storedAppId, deploymentId);
         })
-        .then((packageHistory: Package[]): void => {
-            printDeploymentHistory(command, packageHistory);
+        .then((packageHistory: Package[]): Promise<void> => {
+            return sdk.getDeploymentMetrics(storedAppId, storedDeploymentId)
```
silhouettes: (Open for discussion)

Would it be worth pre-fetching this on the server side to save the round trip? On the server side, you could just call the function that retrieves the metrics directly, and then move the code that merges the results below from the CLI into the server.

cli/script/command-executor.ts
```diff
@@ -254,8 +269,32 @@ export var deploymentList = (command: cli.IDeploymentListCommand, showPackage: b
...
+            
+            var deploymentKeyList: string[];
             return Q.all(deploymentKeyPromises)
-                .then((deploymentKeyList: string[]) => {
+                .then((retrievedDeploymentKeyList: string[]) => {
+                    deploymentKeyList = retrievedDeploymentKeyList;
+                    if (showPackage) {
+                        var metricsPromises: Promise<void>[] = deployments.map((deployment: Deployment) => {
+                            if (deployment.package) {
+                                return sdk.getDeploymentMetrics(theAppId, deployment.id)
```
silhouettes: (Open for discussion) - As with below, would it be worth moving this code onto the server-side to save the round trip?

cli/script/command-executor.ts
geof90: But the metrics route includes metrics for binary app versions as well, not just CodePush packages. Also how would we support the "deployment metrics" command?

cli/script/command-executor.ts
```diff
@@ -657,7 +739,69 @@ function printDeploymentHistory(command: cli.IDeploymentHistoryCommand, packageH
...
+    }
+}
+
+function printDeploymentMetrics(command: cli.IDeploymentMetricsCommand, metrics: any): void {
+    var labelToMetricsMap: { [label: string] : DeploymentMetric } = {};
+    var totalActive: number = 0;
+    var versionRegex = /v(\d+)/;
+    
+    Object.keys(metrics).forEach((metricKey: string) => {
+        var parsedKey: string[] = metricKey.split(":");
```
silhouettes: I feel like the format of the metric key is an implementation detail that should be hidden away on the server - I'm just conscious that this is not the most friendly format as we get other consumers of the SDK. Could we move this parsing onto the server side, and then just return objects that have `.active`, `.downloaded`, etc fields, which can in turn be documented in `rest-definitions.d.ts`? Note we'd likely have to do this anyway if we added this information to the deployment history route.

Sorry to be such a hassle :(

cli/script/command-executor.ts
silhouettes: Seeing as we now have this information, would it be worth displaying the binary app versions in the history as well? I'm not strongly opinionated on this point though, I'm mainly concerned with the round-tripping.

cli/script/command-executor.ts
silhouettes: Or another option is we could move this parsing into the SDK. This makes it less friendly to interact with the API directly, but it means less processing on the server. I'd still lean towards server, but you might have different ideas on the trade-offs (if you agree with me on this point in the first place that is).

geof90:
Richard + Jonathan's comments:

1. Are we going to add the appVersion for each labelled package then? Will this make the sort order look weird (considering you can do something like releasing v1 with appVersion 1.0.0, v2 with 1.0.1, and then v3 back to 1.0.0)?

2. Will display percentages for Active. The concern I had was that the deploymentHistory command does not show metrics for binary app versions, so the percentages may not sum to 100, but i guess that's okay.

3. I prefer to have the percentage beside the Active metric. 

4. What will the success/fail percentages indicate? Success + fail = 100%?

5. Not showing it if 0 and showing it red if >0 makes sense to me.

6. I don't like the term "Users". If I upgraded to a newer version, could I still be a User? It doesnt have the connation of having a particular label it as the current version..

Also to Jonathan's suggestion of displaying installs/downloaded as **Installs: 3,345 (56 pending)** - I feel it is confusing because is the total number of downloads 3345+56 = 3401, or is it 3345, with 3345-56 = 3289 being the number of confirmed installs? I think displaying install and download numbers seperately is clearer.

geof90:
I have made some changes, here's how the display looks now:

screen shot 2016-01-19 at 11 38 13 pm

cli/definitions/cli.ts
```diff
@@ -79,6 +81,12 @@ export interface IDeploymentHistoryCommand extends ICommand {
     format: string;
 }
 
+export interface IDeploymentMetricsCommand extends ICommand {
```
lostintangent: Should this be moved above to be sorted alpha?

cli/script/command-executor.ts
```diff
@@ -676,6 +826,29 @@ function getPackageString(packageObject: Package): string {
...
+    if (packageObject.metrics.installed || numPending) {
+        returnString += "\n" + chalk.green("Installs: ") + ("" + packageObject.metrics.installed);
+    }
+    ;
+    if (numPending) {
+        returnString += " (" + numPending + " pending)";
+    }    
+    
+    if (packageObject.metrics.failed) {
+        returnString += "\n" + chalk.green("Failures: ") + chalk.red(packageObject.metrics.failed + "");
```
lostintangent: What do you think about calling this "Rollbacks" instead? I think it more accurately reflects what it represents since we can't say for sure that it was actually a failure. It could have been a dev error (by not calling notifyApplicarionReady), a timing issue (the device died at the perfect moment), etc.

cli/script/command-executor.ts
```diff
@@ -676,6 +826,29 @@ function getPackageString(packageObject: Package): string {
...
 
+function getPackageMetricsString(packageObject: PackageWithMetrics): string {
+    if (!packageObject || !packageObject.metrics) {
+        return "";
+    }
+    
+    var percentString: string = packageObject.metrics.activePercent === 100.0 ? "100" : packageObject.metrics.activePercent.toPrecision(2) + "%";
+    var numPending: number = packageObject.metrics.downloaded - packageObject.metrics.installed - packageObject.metrics.failed;
+    var returnString: string = chalk.green("Active: ") + percentString + " (" + packageObject.metrics.active + ")";
+    if (packageObject.metrics.installed || numPending) {
```
lostintangent: I'm in favor of showing the installs metric even if it's zero, but only showing the pending number of its non-zero.

cli/script/command-executor.ts
```diff
@@ -657,7 +776,38 @@ function printDeploymentHistory(command: cli.IDeploymentHistoryCommand, packageH
...
+    var labelRegex = /^v\d+$/;
+    var totalActive: number = 0;
+    Object.keys(metrics).forEach((label: string) => {
+        totalActive += metrics[label].active;
+    });
+    
+    if (command.format === "json") {
+        printJson(metrics);
+    } else if (command.format === "table") {
+        printTable(["Label", "Type", "Active", "Downloaded", "Installed", "Failed"], (dataSource: any[]) => {
```
lostintangent: Should we show downloads here? I'd prefer to be consistent with the other commands and display the pending install count instead.

Additionally, instead of adding the "Type" column, what if we added the "App version" column,  and for binary releases, the "Label" column would be empty. If you made the "App version" column the first column, it would act as a sort of "section header" for all of the CodePush releases it contains.

lostintangent:
Should the active percentages display up to two decimal places?

silhouettes:
I think it's unclear that the 'Active' percentage is a percentage of the total number of devices, rather than of the number of installs. Could we just add the total number in the brackets to disambiguate that? e.g. `Active: 8.8% (3 of 33)`. That '33' number would then be the same across all of the versions and hopefully make it deducible.

geof90:
I've removed the "deployment metrics" command and added the total active count in brackets.

lostintangent:
If you can make sure to format the metrics numbers to include commas, then this LGTM!

silhouettes:
Sweet! Looks good

Issue #96, opened 2016-01-06
try-json dependency


cmcewen:
Just a small note, got this warning when installing the cli
 
`npm WARN engine try-json@1.0.0: wanted: {"node":"0.x"} (current: {"node":"5.3.0","npm":"3.3.12"})`

Seems like a pretty simple helper function just to include https://github.com/sorensen/try-json.js/blob/master/index.js

lostintangent:
Yeah, I'm all for not duplicating work, but I agree, there's not a ton of value in us using this library :) I'll remove it and clean up this warning tomorrow. Thanks for reporting this @cmcewen!

lostintangent:
This has been fixed in master and we'll release an update including this to NPM early next week.

PR #111, opened 2016-01-21
Removing try-json dependency

lostintangent:
This simply removes the CLI's dependency on the `try-json` library, which addresses the warning users see when installing the CLI (as reported by #96).

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -38,7 +38,6 @@
     "recursive-fs": "0.1.4",
     "semver": "4.3.6",
     "slash": "1.0.0",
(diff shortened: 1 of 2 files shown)


geof90:
The SDK still uses try-json though. It seems like the warning will appear since the CLI takes a dependency on the SDK.

lostintangent:
Hmm, yeah good point. Well, I'll implement this part to remove the dependency in the CLI and we can leave this bug open to track removing it from the SDK as well (I don't want to make that change right now).

silhouettes:
Looks good!

PR #112, opened 2016-01-21
Fix 100% bug

geof90:
Fixes a tiny bug with the "%" symbol not showing up in the case where the active percentage is 100.


silhouettes:
:+1:

PR #113, opened 2016-01-21
Renaming 'package' parameters

lostintangent:
Replacing uses of the identifier `package` since that is a reserved keyword, which Babel will reject when bundling this file as part of a React Native app. I'm also bumping the version so that we can release this update to NPM.


geof90:
LGTM!

PR #115, opened 2016-01-21
Release table header changes

lostintangent:
This PR simply changes the names of two headers in the deployment and release history, and ensures that the number of pending installs is greater than zero before showing it. For the later, we should investigate why this number would become negative, but in the meantime, it seems semantically correct to explicitly verify that we want a positive number anyways.


geof90:
LGTM!

PR #116, opened 2016-01-21
Display acquisition metrics

geof90:
For the "deployment ls" command, we will show the "Pending" key instead of the "Installed" key by default, since the number of Installed versions for the latest package should be equal to the number of "Active" versions.

Also, if there are no active users, make an exception to show "0%" instead of "0.0%".

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -799,8 +799,8 @@ function getPackageMetricsString(packageObject: PackageWithMetrics, showInstalls
         if (numPending > 0) {
             returnString += " (" + numPending.toLocaleString() + " pending)";
         }
-    } else if (numPending > 0) {
-        returnString += "Pending: " +  numPending.toLocaleString();
+    } else {
+        returnString += "Pending: " + Math.max(numPending, 0).toLocaleString();
     }
     
     if (packageObject.metrics.failed) {
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -719,7 +719,7 @@ function printDeploymentList(command: cli.IDeploymentListCommand, deployments: D
                 
                 if (showPackage) {
                     row.push(getPackageString(deployment.package));
-                    row.push(getPackageMetricsString(<PackageWithMetrics>(deployment.package)));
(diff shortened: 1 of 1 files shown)


cli/script/command-executor.ts
```diff
@@ -782,14 +782,26 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+    
+    var numPending: number = packageObject.metrics.downloaded - packageObject.metrics.installed - packageObject.metrics.failed;
+    var returnString: string = chalk.green("Active: ") + percentString + " (" + packageObject.metrics.active.toLocaleString() + " of " + packageObject.metrics.totalActive.toLocaleString() + ")";
+    if (showInstalls){
+        returnString += "\n" + chalk.green("Installs: ") + packageObject.metrics.installed.toLocaleString();
+        if (numPending > 0) {
+            returnString += " (" + numPending.toLocaleString() + " pending)";
+        }
+    } else if (numPending > 0) {
+        returnString += "Pending: " +  numPending.toLocaleString();
```
lostintangent: I'd be in favor of always showing this number. That way, `deployment ls` always shows the number of active and pending installs, even if one of them is zero. That said, if we always show it, we would just want to not show a negative number, so it might make sense to do `Math.max(numPending, 0).toLocaleString` or something.

cli/script/command-executor.ts
```diff
@@ -782,14 +782,26 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+    
+    var numPending: number = packageObject.metrics.downloaded - packageObject.metrics.installed - packageObject.metrics.failed;
+    var returnString: string = chalk.green("Active: ") + percentString + " (" + packageObject.metrics.active.toLocaleString() + " of " + packageObject.metrics.totalActive.toLocaleString() + ")";
+    if (showInstalls){
+        returnString += "\n" + chalk.green("Installs: ") + packageObject.metrics.installed.toLocaleString();
+        if (numPending > 0) {
+            returnString += " (" + numPending.toLocaleString() + " pending)";
+        }
+    } else {
+        returnString += "Pending: " + Math.max(numPending, 0).toLocaleString();
```
lostintangent: Wouldn't you need a newline at the beginning of this string?

PR #125, opened 2016-01-26
Adding no updates message

lostintangent:
This PR simply adds a new message to the `Update Metadata` column in the table output by `deployment ls` for deployments that haven't had an update released to them yet. It is consistent with the message that we display in the new `Install Metrics` column when a specific deployment/release hasn't had any install reports yet.


shishirx34:
LGTM

PR #126, opened 2016-01-27
Metrics SDK Changes

geof90:
Change the Acquisition SDK to allow clients to send previous version deployment key and label as part of the client state.

### DIFF
--- a/definitions/rest-definitions.d.ts
+++ b/definitions/rest-definitions.d.ts
@@ -15,8 +15,8 @@ declare module "rest-definitions" {
         appVersion: string;
         clientUniqueId?: string;
         deploymentKey: string;
-        fromDeploymentKey?: string;
-        fromLabelOrAppVersion?: string;
+        previousDeploymentKey?: string;
+        previousLabelOrAppVersion?: string;
         label?: string;
         status?: string;
     }
--- a/definitions/rest-definitions.d.ts
+++ b/definitions/rest-definitions.d.ts
@@ -13,10 +13,12 @@ declare module "rest-definitions" {
 
     export interface DeploymentStatusReport {
         appVersion: string;
-        clientUniqueId: string;
-        deploymentKey: string; 
+        clientUniqueId?: string;
+        deploymentKey: string;
+        fromDeploymentKey?: string;
+        fromLabelOrAppVersion?: string;
         label?: string;
-        status?: string
+        status?: string;
     }
     
     export interface DownloadReport {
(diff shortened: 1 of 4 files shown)


definitions/rest-definitions.d.ts
```diff
@@ -13,10 +13,12 @@ declare module "rest-definitions" {
 
     export interface DeploymentStatusReport {
         appVersion: string;
-        clientUniqueId: string;
-        deploymentKey: string; 
+        clientUniqueId?: string;
+        deploymentKey: string;
+        fromDeploymentKey?: string;
```
lostintangent: How about "sourceDeploymentKey" and "sourceLabelOrAppVersion"?

sdk/script/acquisition-sdk.ts
```diff
@@ -169,7 +172,17 @@ export class AcquisitionManager {
                     return;
             }
         }
-
+        
+        if (fromLabelOrAppVersion) {
```
lostintangent: Do we need to make this check since setting the property to undefined/null would have the affect of not creating the property at all.

definitions/rest-definitions.d.ts
geof90: I like "previous" more than "source", the "source" is the server. I'll change it all to "previous" then.

sdk/script/acquisition-sdk.ts
geof90: I'm not sure about `undefined`, but `null` would indeed create the property. Sure the server ignores it if it is falsey, but if we can save the network overhead of sending few bytes from the client side I think it makes sense to do it.

lostintangent:
LGTM!

definitions/rest-definitions.d.ts
```diff
@@ -13,10 +13,12 @@ declare module "rest-definitions" {
 
     export interface DeploymentStatusReport {
         appVersion: string;
-        clientUniqueId: string;
-        deploymentKey: string; 
+        clientUniqueId?: string;
+        deploymentKey: string;
+        previousDeploymentKey?: string;
+        previousLabelOrAppVersion?: string;
```
shishirx34: may be create an object to aggregate these properties? If you think it makes sense.
```
export interface PreviousAppConfiguration
   {
        previoustDeploymentKey?: string;
        previousLabelOrAppVersion?: string;
   }
```

shishirx34:
LGTM

definitions/rest-definitions.d.ts
geof90: I think this makes sense if we are going to use these properties somewhere else, but for now, I can't forsee this ever being used outside of reporting deployment status updates to a new version. As such, I think I prefer to keep it flat.

silhouettes:
Looks good~

PR #128, opened 2016-02-03
Tidy up SDK

silhouettes:
This was originally intended to be done later as part of productizing the SDK, however this quick PR simplifies my other work too.

Changes:

1. Clean up/de-deduplicate all of the boilerplate. The number of lines is halved and it's much more readable. I've also found out that we have some inconsistencies with how we handle response bodies in POST requests.
2. Convert everything to use template strings
3. Remove try-json dependency

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -18,7 +18,6 @@
     "fs": "0.0.2",
     "node-uuid": "^1.4.3",
     "q": "^1.4.1",
-    "superagent": "git://github.com/visionmedia/superagent.git#45a1290b2fe5a56a1d77ca74e0b236178d58d848",
-    "try-json": "^1.0.0"
+    "superagent": "git://github.com/visionmedia/superagent.git#45a1290b2fe5a56a1d77ca74e0b236178d58d848"
   }
 }
--- a/sdk/script/account-manager.ts
+++ b/sdk/script/account-manager.ts
@@ -59,7 +59,7 @@ export class AccountManager {
 
     public isAuthenticated(): Promise<boolean> {
         return Promise<boolean>((resolve, reject, notify) => {
-            var request: superagent.Request<any> = superagent.get(this._serverUrl + `/authenticated`);
+            var request: superagent.Request<any> = superagent.get(`${this._serverUrl}/authenticated`);
             this.attachCredentials(request);
 
             request.end((err: any, res: superagent.Response) => {
@@ -82,7 +82,7 @@ export class AccountManager {
                 var accessKey: AccessKey = { id: null, name: newAccessKey, createdTime: new Date().getTime(), createdBy: machine, description: description };
                 return this.post(`/accessKeys/`, JSON.stringify(accessKey), /*expectResponseBody=*/ false)
                     .then((res: JsonResponse) => {
-                        var location = res.header[`location`];
+                        var location: string = res.header[`location`];
                         if (location && location.lastIndexOf(`/`) !== -1) {
                             accessKey.id = location.substr(location.lastIndexOf(`/`) + 1);
                             return accessKey;
@@ -131,7 +131,7 @@ export class AccountManager {
     }
 
     public addApp(appName: string): Promise<App> {
-        var app = <App>{ name: appName };
+        var app: App = { name: appName };
         return this.post(`/apps/`, JSON.stringify(app), /*expectResponseBody=*/ false)
             .then((res: JsonResponse) => {
                 var location = res.header[`location`];
@@ -145,7 +145,7 @@ export class AccountManager {
     }
 
     public removeApp(app: App | string): Promise<void> {
-        var id: string = (typeof app === `string`) ? (<string><any>app) : (<App><any>app).id;
+        var id: string = (typeof app === `string`) ? (<string>app) : (<App>app).id;
         return this.del(`/apps/${id}`)
             .then(() => null);
     }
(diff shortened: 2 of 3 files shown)


sdk/package.json
```diff
@@ -18,7 +18,6 @@
     "fs": "0.0.2",
     "node-uuid": "^1.4.3",
     "q": "^1.4.1",
-    "superagent": "git://github.com/visionmedia/superagent.git#45a1290b2fe5a56a1d77ca74e0b236178d58d848",
-    "try-json": "^1.0.0"
+    "superagent": "git://github.com/visionmedia/superagent.git#45a1290b2fe5a56a1d77ca74e0b236178d58d848"
```
geof90: https://github.com/visionmedia/superagent/pull/771 was merged and published to NPM, should we depend on the latest NPM version instead?

sdk/script/account-manager.ts
```diff
@@ -9,12 +8,12 @@ var packageJson = require("../../package.json");
 
 declare var fs: any;
 
-if (typeof window === "undefined") {
-    fs = require("fs");
+if (typeof window === `undefined`) {
```
geof90: I'm just wondering whether we should from now on use template strings for everything, or only when it is required.

sdk/script/account-manager.ts
```diff
@@ -55,7 +59,7 @@ export class AccountManager {
 
     public isAuthenticated(): Promise<boolean> {
         return Promise<boolean>((resolve, reject, notify) => {
-            var request: superagent.Request<any> = superagent.get(this._serverUrl + "/authenticated");
+            var request: superagent.Request<any> = superagent.get(this._serverUrl + `/authenticated`);
```
geof90: Why not `${this._serverUrl}/authenticated?`

sdk/script/account-manager.ts
silhouettes: I like the idea of using it for everything, but what do you think?

The only tricky thing is with my version of the TS compiler, it doesn't seem to like it when you use template strings with ES6 style import statements.

sdk/package.json
silhouettes: Yes, although it might be good to do it on a non-breaking branch, as this one won't go in for a while.

silhouettes: Ehhhh, I can just do it here I guess

sdk/script/account-manager.ts
```diff
@@ -73,523 +77,133 @@ export class AccountManager {
...
-                            var location = res.header["location"];
-                            if (location && location.lastIndexOf("/") !== -1) {
-                                accessKey.id = location.substr(location.lastIndexOf("/") + 1);
-                                resolve(accessKey);
-                            } else {
-                                resolve(null);
-                            }
+                return this.post(`/accessKeys/`, JSON.stringify(accessKey), /*expectResponseBody=*/ false)
+                    .then((res: JsonResponse) => {
+                        var location = res.header[`location`];
```
geof90: : string?

geof90: : string?

lostintangent:
LGTM! This is an amazing clean up job :)

sdk/script/account-manager.ts
```diff
@@ -73,523 +77,133 @@ export class AccountManager {
...
 
-                var body = tryJSON(res.text);
-                if (res.ok) {
-                    if (body) {
-                        resolve(body.app);
-                    } else {
-                        reject(<CodePushError>{ message: "Could not parse response: " + res.text, statusCode: res.status });
-                    }
+    public addApp(appName: string): Promise<App> {
+        var app = <App>{ name: appName };
```
geof90: why not `var app: App = { name: appName };` ?

sdk/script/account-manager.ts
```diff
@@ -73,523 +77,133 @@ export class AccountManager {
...
-                    var body = tryJSON(res.text);
-                    if (body) {
-                        reject(<CodePushError>body);
-                    } else {
-                        reject(<CodePushError>{ message: res.text, statusCode: res.status });
-                    }
-                }
-            });
-        });
+        var id: string = (typeof app === `string`) ? (<string><any>app) : (<App><any>app).id;
```
geof90: Is there a reason why this double casting is needed? Why can't we just do <string>app or <App>app?

sdk/script/account-manager.ts
```diff
@@ -73,523 +77,133 @@ export class AccountManager {
...
-                    var body = tryJSON(res.text);
-                    if (body) {
-                        reject(<CodePushError>body);
-                    } else {
-                        reject(<CodePushError>{ message: res.text, statusCode: res.status });
-                    }
-                }
-            });
-        });
+        var id: string = (typeof deployment === `string`) ? (<string><any>deployment) : (<Deployment><any>deployment).id;
```
geof90: Same question as above about the double casting

geof90:
LGTM!

sdk/script/account-manager.ts
silhouettes: Weird. That wasn't compiling before but it does now (which makes sense). Changed it to the more logical option here

sdk/script/account-manager.ts
dtivel: I don't think we should use template strings for everything.  As with other language features, we should use template strings when their use is appropriate and their benefits are realized.

dtivel:
Impressive simplification.  Looks good.

PR #133, opened 2016-02-13
Move name resolution logic out of the CLI

silhouettes:
(Merging into 'breaking'. On the diff tab, add '?w=1' to the URL to ignore whitespace changes)

Note that the issue of disambiguating duplicate apps by email is not explicitly addressed here. I will address this in a subsequent review, but I expect that it will be a matter of URL encoding the '<email>/appName' chunk and then decoding it on the server.

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -18,7 +18,7 @@ import wordwrap = require("wordwrap");
 
 import * as cli from "../definitions/cli";
 import { AcquisitionStatus } from "code-push/script/acquisition-sdk";
-import { AccessKey, AccountManager, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Package, Permissions, UpdateMetrics } from "code-push";
+import { AccessKey, AccountManager, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Package, Permissions, UpdateMetrics } from "code-push";
 var packageJson = require("../package.json");
 import Promise = Q.Promise;
 var progress = require("progress");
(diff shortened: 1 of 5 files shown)


cli/script/command-executor.ts
```diff
@@ -1180,48 +939,12 @@ function isBinaryOrZip(path: string): boolean {
...
-    if (!providerUniqueId) throw new Error("Provider unique ID is missing.");
-}
-
-function throwForInvalidAccessKeyId(accessKeyId: string, accessKeyName: string): void {
-    if (!accessKeyId) {
-        throw new Error("Access key \"" + accessKeyName + "\" does not exist.");
-    }
-}
-
-function throwForInvalidApp(app: App, appName: string): void {
```
silhouettes: Verify that error messages are unchanged - as we are resolving names on the server, an invalid name will result in an error coming from the server rather than the CLI.

sdk/script/account-manager.ts
```diff
@@ -367,7 +346,7 @@ export class AccountManager {
                     .toString(`base64`)
                     .replace(/\+/g, `_`)  // URL-friendly characters
                     .replace(/\//g, `-`)
-                    .concat(account.id);
+                    .concat(account.email);
```
silhouettes: I will address this in a different review - For now, I expect to move the access-key generation onto the server.

cli/script/command-executor.ts
```diff
@@ -18,7 +18,7 @@ import wordwrap = require("wordwrap");
 
 import * as cli from "../definitions/cli";
 import { AcquisitionStatus } from "code-push/script/acquisition-sdk";
-import { AccessKey, AccountManager, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Package, Permissions, UpdateMetrics } from "code-push";
+import { AccessKey, AccountManager, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Package, Permissions, UpdateMetrics } from "code-push";
```
dtivel: Account comes before AccountManager

dtivel:
Looks good.

cli/script/command-executor.ts
silhouettes: I will need to add this as part of a server change - currently we just get: `[Error]  Not Found`

PR #134, opened 2016-02-13
release-react command

geof90:
Will update docs in a seperate PR.

screen shot 2016-02-12 at 8 07 38 pm


cli/script/command-executor.ts
```diff
@@ -1174,6 +1256,69 @@ function release(command: cli.IReleaseCommand): Promise<void> {
...
+        }
+    } else {
+        if (fileDoesNotExistOrIsDirectory(entryFile)) {
+            throw new Error(`Entry file "${entryFile}" does not exist.`);
+        }
+    }
+    
+    // This is needed to clear the react native bundler cache:
+    // https://github.com/facebook/react-native/issues/4289
+    execSync("rm -rf $TMPDIR/react-*");
```
lostintangent: Shouldn't we use the rimraf library that you added a dependency for? Would this work correctly on Windows?

cli/script/command-executor.ts
```diff
@@ -1223,6 +1364,43 @@ function requestAccessToken(): Promise<string> {
     });
 }
 
+export var runReactNativeBundleCommand = (entryFile: string, outputFolder: string, platform: string, sourcemapOutput: string): Promise<void> => {
+    var reactNativeBundleCommandArgs = [
```
lostintangent: I wonder if we should use the locally installed packager to run this, as opposed to relying on the global CLI. That way we know which versions of RN we support since we'd be depending on a Node module that's shipped with the platform as opposed to a globally installed CLI. It seems a little cleaner to use a n NPM module instead of shelling to a CLI if possible as well.

https://github.com/facebook/react-native/blob/master/packager/react-packager/index.js

cli/script/command-executor.ts
```diff
@@ -1020,6 +1057,50 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+            throw new Error("Unable to parse the \"CFBundleShortVersionString\" from \"Info.plist\".");
+        }
+    } else {
+        try {
+            var buildGradleContents: string = fs.readFileSync(path.join("android", "app", "build.gradle")).toString();
+        } catch (err) {
+            throw new Error("Unable to find or read \"build.gradle\" in the \"android/app\" folder.");
+        }
+        
+        var appVersionRegex: RegExp = /versionName\s+"(.*?)"/;
```
lostintangent: I'm a little weary of this since it seems pretty possible for this regex to match multiple times within a single build.gradle file. I wonder whether it wouldn't be a better idea to try to read the version from the AndroidManifest.xml that is ultimately generated from your gradle build. Parsing XML seems like a much better option than Groovy.

cli/package.json
```diff
@@ -37,11 +37,15 @@
...
+    "rimraf": "^2.5.1",
     "semver": "4.3.6",
     "slash": "1.0.0",
     "update-notifier": "^0.5.0",
     "wordwrap": "1.0.0",
     "yargs": "^3.15.0",
     "yazl": "2.2.2"
+  },
+  "peerDependencies": {
+    "react-native-cli": "^0.1.10"
```
lostintangent: This would result in a warning for Cordova users that don't have RN installed? If so, I think we should skip adding it, and instead rely on dynamically requiring the packager at runtime like I mentioned in my other comment.

cli/script/command-executor.ts
```diff
@@ -59,8 +62,12 @@ export interface PackageWithMetrics {
     metrics?: UpdateMetricsWithTotalActive;
 }
 
-export var sdk: AccountManager;
 export var log = (message: string | Chalk.ChalkChain): void => console.log(message);
+export var sdk: AccountManager;
+export var spawn = childProcess.spawn;
+export var getTmpDir = (): string => {
```
lostintangent: I almost think it would be clearer to just use "os.tmpdir()" below instead of wrapping that expression in a method. If you wanted just a method, could t you do something like "var {tmpdir} = os"?

geof90: This was done so that I could override the method with a stub during tests.

cli/script/command-executor.ts
geof90: Is there a possibility where a build.gradle file will contain multiple "versionName" fields?

lostintangent: Sure. In practice, it might not be common, but any number of other plugins or config objects could also have a method call versionName. Additionally, the value being passed to the versionName method could also be dynamic, and therefore, simply capturing the string like this wouldn't be sufficient. Once again, that could be rare, but we already know of one example of a native Android app integrating RN into it, and we should keep that in mind here. The AndroidManifest.xml solution wouldn't have either of these issues, and might merits a quick investigation.

cli/script/command-executor.ts
geof90: Controlling the packager directly seems error prone.. We'd be duplicating logic from `react-native/local-cli`. Could we just take a direct dependency on `react-native-cli` and use node to run it like how we did for the vsts task?

cli/script/command-executor.ts
geof90: The AndroidManifest.xml template that RN generates does not contain the app version.

cli/script/command-executor.ts
lostintangent: Ah OK. It might be worth looking into Jest as a test runner at some point. It auto mocks required modules (with the ability to fake their behavior), which makes it pretty clean to write unit tests, without needing to abstract existing modules any further simply to be able to mock/stub them within tests.

geof90: Actually.. maybe its not necessary to stub this. I was initially worried that os.tmpdir() might not return a constant value.

cli/script/command-executor.ts
lostintangent: Ah, I forgot that the local CLI does a bunch of work (the relationship between the global CLI, local CLI and packager is pretty confusing), so you're totally right that we shouldn't consume the packager for this.

If we consumed the local CLI directly would we need to duplicate a bunch of work in the global CLI? If we absolutely have to use the global CLI, then that's cool, I'm just curious if that is the case.

cli/script/command-executor.ts
geof90: If we parsed this in a smarter way, e.g. make sure we are looking at android.defaultConfig.versionName, would it avoid the scenario where other plugin or config objects have the versionName method?

Granted, that wouldn't handle the case where it is a dynamic value, e.g `getVersionName()`. If its a rare enough case, I guess its okay to allow those users to fall back to the original way of manually specifying the appVersion to our CLI?

cli/script/command-executor.ts
geof90: On a quick glance... using local-cli seems possible. I'll need to investigate it further, but all the global-cli seems to do in the case of the `react-native bundle` command is try to find the path to the local CLI and require it. It does a lot more e.g includes a lot of troubleshooting information for the user in the case the user is running "react-native init", but that's not what we care about.

geof90: Using local-cli works like a charm!!

lostintangent: Awesome!

cli/script/command-executor.ts
lostintangent: I would be cool with that if we made sure to catch the right versionName.

cli/script/command-executor.ts
```diff
@@ -2,14 +2,18 @@
 
 import * as base64 from "base-64";
 import * as chalk from "chalk";
+var childProcess = require("child_process");
 import * as fs from "fs";
```
shishirx34: you removed `fs` package from package.json and still are using here. I suppose you didn't clean your binaries that is why it worked fine for you, however this would fail, wouldn't it? Is there someplace else where we have this in our project?

cli/script/command-executor.ts
```diff
@@ -229,6 +234,13 @@ function appRename(command: cli.IAppRenameCommand): Promise<void> {
         });
 }
 
+export var createEmptyTempReleaseFolder = (folderPath: string) => {
+    return deleteFolder(folderPath)
+        .then(() => {
+            fs.mkdirSync(folderPath);
```
shishirx34: is `fs` module correctly included in the project?

cli/script/command-parser.ts
```diff
@@ -207,6 +207,19 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
 
         addCommonConfiguration(yargs);
     })
+    .command("release-react", "Release a new version of your React Native app to a specific deployment", (yargs: yargs.Argv) => {
+        yargs.usage(USAGE_PREFIX + " release-react <appName> <platform> [--deploymentName <deploymentName>] [--description <description>] [--entryFile <entryFile>] [--mandatory] [--sourcemapOutput <sourcemapOutput>]")
+            .demand(/*count*/ 3, /*max*/ 3)  // Require exactly four non-option arguments.
```
dtivel: three

cli/script/command-executor.ts
```diff
@@ -1223,6 +1362,43 @@ function requestAccessToken(): Promise<string> {
...
+        "--entry-file", entryFile,
+        "--platform", platform,
+    ];
+    
+    if (sourcemapOutput) {
+        reactNativeBundleArgs.push("--sourcemap-output", sourcemapOutput);
+    }
+    
+    log(chalk.cyan("Running \"react-native bundle\" command:\n"));
+    var reactNativeBundleCommand = spawn("node", reactNativeBundleArgs);
```
dtivel: Maybe `Process` would be better than `Command`.

cli/script/command-executor.ts
```diff
@@ -1020,6 +1055,50 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+            } catch (err) {
+                throw new Error(`Unable to find or read "Info.plist" in the "iOS/${projectName}" or "iOS" folders.`);
+            }
+        }
+        
+        var appVersionRegex: RegExp = /CFBundleShortVersionString\s*<\s*\/key\s*>\s*<\s*string\s*>(.*?)<\s*\/string\s*>/;
+        var parsedInfoPlist: string[] = infoPlistContents.match(appVersionRegex);
+        if (parsedInfoPlist && parsedInfoPlist[1]) {
+             if (semver.valid(parsedInfoPlist[1]) === null) {
+                throw new Error("Please update \"Info.plist\" to use a semver compliant \"CFBundleShortVersionString\", for example \"1.0.3\".");
```
dtivel: semver-compliant

cli/script/command-executor.ts
```diff
@@ -1020,6 +1055,50 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+        }
+        
+        return g2js.parseFile(buildGradlePath)
+            .catch((err: Error) => {
+                throw new Error("Unable to parse the \"versionName\" from \"build.gradle\".");
+            })
+            .then((buildGradle: any) => {
+                var appVersion: string = buildGradle.android.defaultConfig.versionName.replace(/"/g, "").trim();
+                if (semver.valid(appVersion) === null) {
+                    throw new Error("Please update \"build.gradle\" to use a semver compliant \"versionName\", for example \"1.0.3\".");
```
dtivel: semver-compliant

cli/script/command-executor.ts
```diff
@@ -1084,7 +1163,7 @@ function promote(command: cli.IPromoteCommand): Promise<void> {
         });
 }
 
-function release(command: cli.IReleaseCommand): Promise<void> {
+export var release = (command: cli.IReleaseCommand): Promise<void> => {
```
dtivel: semver-compliant (a few lines below) and also in cli/README.md.

cli/script/command-executor.ts
```diff
@@ -1174,6 +1253,66 @@ function release(command: cli.IReleaseCommand): Promise<void> {
         });
 }
 
+export var releaseReact = (command: cli.IReleaseReactCommand): Promise<void> => {
+    var platform: string = command.platform.toLowerCase();
+    var entryFile: string = command.entryFile;
+    var outputFolder: string = path.join(os.tmpdir(), "CodePush");
+    var releaseCommand: cli.IReleaseCommand = <any>command;
```
dtivel: This <any> cast feels a little fragile because common members between the two interfaces (IReleaseCommand and IReleaseReactCommand) are expected to stay in sync.  Perhaps a common base interface or a comment (at the interface definition sites) would help.

dtivel:
Looks good.

cli/script/command-executor.ts
```diff
@@ -1020,6 +1055,50 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+        } catch (err) {
+            try {
+                infoPlistContents = fs.readFileSync(path.join("iOS", "Info.plist")).toString();
+            } catch (err) {
+                throw new Error(`Unable to find or read "Info.plist" in the "iOS/${projectName}" or "iOS" folders.`);
+            }
+        }
+        
+        var appVersionRegex: RegExp = /CFBundleShortVersionString\s*<\s*\/key\s*>\s*<\s*string\s*>(.*?)<\s*\/string\s*>/;
+        var parsedInfoPlist: string[] = infoPlistContents.match(appVersionRegex);
```
shishirx34: Isn't there an xml node module that we can use to parse the appVersion? Are you fairly confident that this RegEx won't break down the line if/when the format for info.plist changes?

cli/script/command-executor.ts
geof90: I have added a base interface `IReleaseBaseCommand`.

cli/script/command-executor.ts
geof90: Added https://www.npmjs.com/package/plist

cli/script/command-executor.ts
```diff
@@ -1020,6 +1056,49 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+                infoPlistContents = fs.readFileSync(path.join("iOS", "Info.plist")).toString();
+            } catch (err) {
+                throw new Error(`Unable to find or read "Info.plist" in the "iOS/${projectName}" or "iOS" folders.`);
+            }
+        }
+        
+        var parsedInfoPlist: any = plist.parse(infoPlistContents);
+        if (parsedInfoPlist && parsedInfoPlist.CFBundleShortVersionString) {
+             if (semver.valid(parsedInfoPlist.CFBundleShortVersionString) === null) {
+                throw new Error("Please update \"Info.plist\" to use a semver-compliant \"CFBundleShortVersionString\", for example \"1.0.3\".");
```
shishirx34: @lostintangent - should we just keep 1.0.0 as the default value in example/error/help messages?

cli/script/command-executor.ts
```diff
@@ -1020,6 +1056,49 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+        }
+        
+        return g2js.parseFile(buildGradlePath)
+            .catch((err: Error) => {
+                throw new Error("Unable to parse the \"versionName\" from \"build.gradle\".");
+            })
+            .then((buildGradle: any) => {
+                var appVersion: string = buildGradle.android.defaultConfig.versionName.replace(/"/g, "").trim();
+                if (semver.valid(appVersion) === null) {
+                    throw new Error("Please update \"build.gradle\" to use a semver-compliant \"versionName\", for example \"1.0.3\".");
```
shishirx34: `1.0.0` if @lostintangent agrees to default it.

cli/script/command-executor.ts
```diff
@@ -1020,6 +1056,49 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
...
+        if (fileDoesNotExistOrIsDirectory(buildGradlePath)) {
+            throw new Error("Unable to find or read \"build.gradle\" in the \"android/app\" folder.");
+        }
+        
+        return g2js.parseFile(buildGradlePath)
+            .catch((err: Error) => {
+                throw new Error("Unable to parse the \"versionName\" from \"build.gradle\".");
+            })
+            .then((buildGradle: any) => {
+                var appVersion: string = buildGradle.android.defaultConfig.versionName.replace(/"/g, "").trim();
```
shishirx34: This might throw undefined error if any of the properties are missing `android` `defaultConfig` `versionName`, may be add an exist check for those?

cli/script/command-executor.ts
```diff
@@ -1174,6 +1253,66 @@ function release(command: cli.IReleaseCommand): Promise<void> {
...
+    var outputFolder: string = path.join(os.tmpdir(), "CodePush");
+    var releaseCommand: cli.IReleaseCommand = <any>command;
+    releaseCommand.package = outputFolder;
+    
+    if (platform !== "ios" && platform !== "android") {
+        throw new Error("Platform must be either \"ios\" or \"android\".");
+    }
+    
+    try {
+        var projectPackageJson = require(path.join(process.cwd(), "package.json"));
```
shishirx34: : any

cli/script/command-executor.ts
```diff
@@ -1223,6 +1362,43 @@ function requestAccessToken(): Promise<string> {
     });
 }
 
+export var runReactNativeBundleCommand = (entryFile: string, outputFolder: string, platform: string, sourcemapOutput: string): Promise<void> => {
+    var reactNativeBundleArgs = [
+        path.join("node_modules", "react-native", "local-cli", "cli.js"), "bundle",
+        "--assets-dest", outputFolder,
+        "--bundle-output", path.join(outputFolder, "main.jsbundle"),
+        "--dev", false,
```
shishirx34: is the default value for this flag `true` for `react-native` command? If not remove it.

shishirx34:
LGTM!

cli/script/command-executor.ts
geof90: Yes it is.

PR #136, opened 2016-02-18
Breaking changes loose ends

silhouettes:
(Merging into 'breaking')

These are a number of small tweaks that need to be made before the breaking changes branch is merged:

- URL encode all REST API inputs (for example, if we want to pass a qualified app name such as `owner@email.com/app`)
- Move access-key generation onto the server
- Reflect structural API changes (PUT requests are now PATCH requests, the /package endpoint is now renamed to /release)

I will be doing a lot of validation today and also be submitting an (optional) PR with the amalgamated collection of breaking changes to merge into master.

### DIFF
--- a/sdk/script/account-manager.ts
+++ b/sdk/script/account-manager.ts
@@ -54,7 +54,7 @@ export class AccountManager {
 
     constructor(accessKey: string, userAgent?: string, serverUrl?: string) {
         this._accessKey = accessKey;
-        this._userAgent = userAgent || (`${packageJson.name}/${packageJson.version}`);
+        this._userAgent = userAgent;
         this._serverUrl = serverUrl || `https://codepush.azurewebsites.net`;
     }
 
@@ -334,9 +334,12 @@ export class AccountManager {
     }
 
     private attachCredentials(request: superagent.Request<any>): void {
-        request.set(`User-Agent`, this._userAgent);
         request.set(`Accept`, `application/vnd.code-push.${AccountManager.API_VERSION}+json`);
         request.set(`Authorization`, `Bearer ${this._accessKey}`);
+        if (this._userAgent) {
+            request.set(`User-Agent`, this._userAgent);
+        }
+        request.set(`X-CodePush-SDK-Version`, `${packageJson.name}/${packageJson.version}`);
     }
 
     private generateAccessKey(): Promise<string> {
--- a/sdk/script/account-manager.ts
+++ b/sdk/script/account-manager.ts
@@ -45,6 +45,19 @@ interface JsonResponse {
     body?: any;
 }
 
+// A template string tag function that URL encodes the substituted values
+function urlEncode(strings: string[], ...values: string[]): string {
+    var result = "";
+    for (var i = 0; i < strings.length; i++) {
+        result += strings[i];   // Assumes that template string will not start with a substitution
+        if (i < values.length) {
+            result += encodeURIComponent(values[i]);
+        }
+    }
+
+    return result;
+}
+
 export class AccountManager {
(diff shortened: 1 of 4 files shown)


sdk/script/account-manager.ts
```diff
@@ -45,6 +45,19 @@ interface JsonResponse {
     body?: any;
 }
 
+// A template string tag function that URL encodes the substituted values
+function urlEncode(strings: string[], ...values: string[]): string {
+    var result = "";
+    for (var i = 0; i < strings.length; i++) {
+        result += strings[i];   // Assumes that template string will not start with a substitution
```
geof90: From what I experimented with template strings,
```javascript
tag`${1}abc${2}${3}`
```
will yield to the tag function 
```
strings: ["", "def", "", ""]
values: [1, 2, 3]
```

I think the spec ensures that the start will be an empty string, and that contiguous substitutions will also be delimited with an empty string. So the assumption is not needed, it is guaranteed by the spec.

silhouettes: Oh!! Boy am I glad to hear that. I was feeling like this language feature was badly designed, but what you just explained makes perfect sense.

sdk/script/account-manager.ts
```diff
@@ -334,21 +339,11 @@ export class AccountManager {
...
-                    .replace(/\+/g, `_`)  // URL-friendly characters
-                    .replace(/\//g, `-`)
-                    .concat(account.email);
-
-                return accessKey;
-            });
+        if (this._userAgent) {
+            request.set(`User-Agent`, this._userAgent);
+        }
+        request.set(`X-CodePush-SDK-Version`, `${packageJson.name}/${packageJson.version}`);
```
geof90: We don't attach the name here, just the version:

https://github.com/Microsoft/react-native-code-push/blob/master/request-fetch-adapter.js#L11

Perhaps we should standardize so it is easier manage on the server.

sdk/script/account-manager.ts
```diff
@@ -82,124 +95,121 @@ export class AccountManager {
...
-                    .then(() => accessKey);
-            });
+        var accessKeyRequest: AccessKey = { createdBy: machine, description: description };
+        return this.post(urlEncode `/accessKeys/`, JSON.stringify(accessKeyRequest), /*expectResponseBody=*/ true)
+            .then((response: JsonResponse) => response.body.accessKey);
     }
 
     public getAccessKey(accessKey: string): Promise<AccessKey> {
-        return this.get(`/accessKeys/${accessKey}`)
+        return this.get(urlEncode `/accessKeys/${accessKey}`)
```
geof90: Should we just urlEncode inside the `.get(), .post()... ` etc methods? That way we don't have to attach the tag everywhere.

geof90: Oh I see. We can't because those methods will see the compiled template string. Ignore my comment!

sdk/script/account-manager.ts
```diff
@@ -334,21 +339,11 @@ export class AccountManager {
...
-                var accessKey = crypto.randomBytes(21)
-                    .toString(`base64`)
-                    .replace(/\+/g, `_`)  // URL-friendly characters
-                    .replace(/\//g, `-`)
-                    .concat(account.email);
-
-                return accessKey;
-            });
+        if (this._userAgent) {
+            request.set(`User-Agent`, this._userAgent);
```
dtivel: Above you have great examples of template literals used effectively.  However, if string interpolation is not required, use plain old string literals.  This feedback applies globally.

dtivel:
Looks good.

PR #137, opened 2016-02-19
Sorting command list

lostintangent:
This simply sorts the list of commands so that the output of the `code-push` command are displayed appropriately.


dtivel:
Looks good.

PR #138, opened 2016-02-19
Rely on server for generation of qualified app names

silhouettes:
(Merging into 'breaking')

The server now accepts qualified app names as input and also sends them as output where appropriate, so that this logic does not need to be performed by clients.

This will be the last set of breaking changes on the client (I didn't realize I needed to do this when I sent the wrap-up PR yesterday)

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -598,6 +598,7 @@ function printAppList(format: string, apps: App[], deploymentLists: string[][]):
         var dataSource: any[] = apps.map((app: App, index: number) => {
             var augmentedApp: any = app;
             augmentedApp.deployments = deploymentLists[index];
+            return augmentedApp;
         });
 
         printJson(dataSource);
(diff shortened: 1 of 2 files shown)


shishirx34:
LGTM

cli/script/command-executor.ts
```diff
@@ -614,35 +593,11 @@ function formatDate(unixOffset: number): string {
...
-}
-
 function printAppList(format: string, apps: App[], deploymentLists: string[][]): void {
-    var appNameToCountMap: NameToCountMap = getNameToCountMap(apps);
-
     if (format === "json") {
         var dataSource: any[] = apps.map((app: App, index: number) => {
-            return { "name": getAppDisplayName(app, appNameToCountMap), "deployments": deploymentLists[index] };
+            var augmentedApp: any = app;
+            augmentedApp.deployments = deploymentLists[index];
```
geof90: Dont you need to return something?

geof90:
LGTM otherwise!

cli/script/command-executor.ts
silhouettes: Thanks for catching this, big oversight on my part there.

PR #140, opened 2016-02-22
Reflect breaking server changes on the client

silhouettes:
(Add ?w=1 to the URL on the diff view to ignore whitespace changes)

This is an optional PR in preparation for the upcoming breaking server changes. Each of the following pieces have been individually code reviewed, but the merging of features from master has not. Any comments, or especially help with validation, would be very much appreciated.

1. **Cookie-less/session-less authentication.** Instead of retrieving a cookie from the `/auth/login/accessKey` endpoint, we send the access key with each request using the `Authorization` header, similarly to how OAuth works.
1. **Use names instead of ID's at REST endpoints.** Basically moved name resolution logic from the CLI into the server.
1. **Adopt other API simplifications**. Change all PUT routes to PATCH routes. Add 'key' property to Deployment object and remove DeploymentKey endpoints. Change PUT `/package` route to POST `/release`.
1. **SDK cleanup**. Including significant refactoring, and URL encoding all REST inputs.


sdk/script/account-manager.ts
```diff
@@ -0,0 +1,349 @@
...
+    private static API_VERSION: number = 2;
+
+    private _accessKey: string;
+    private _serverUrl: string;
+    private _userAgent: string;
+
+    constructor(accessKey: string, userAgent?: string, serverUrl?: string) {
+        this._accessKey = accessKey;
+        this._userAgent = userAgent;
+        this._serverUrl = serverUrl || "https://codepush.azurewebsites.net";
```
silhouettes: Change this default to `https://codepush-management.azurewebsites.net` to reflect the splitting of the management server from the acquisition server.

sdk/script/account-manager.ts
```diff
@@ -0,0 +1,349 @@
+import * as base64 from "base-64";
```
silhouettes: Other requested SDK improvements, such as making AccountManager the default export, and renaming all instances of 'Package' to 'Update', will be addressed in a subsequent PR.

cli/script/command-executor.ts
```diff
@@ -499,30 +363,31 @@ function deserializeConnectionInfo(): IStandardLoginConnectionInfo | IAccessKeyL
...
-            return logout(<cli.ILogoutCommand>command);
+                    if (!connectionInfo) {
+                        throw new Error("You are not currently logged in. Run the 'code-push login' command to authenticate with the CodePush server.");
+                    }
 
-        case cli.CommandType.register:
-            return register(<cli.IRegisterCommand>command);
-    }
+                    var accessKey: string = getAccessKeyFromConnectionInfo(connectionInfo);
+                    sdk = new AccountManager(accessKey, userAgent, connectionInfo.serverUrl);
```
silhouettes: Need to address the fact that people with older connectionInfo files sitting on disk may still authenticate to the old acquisition server instead of the new management one, even with an upgraded CLI.

PR #144, opened 2016-02-27
[release-react] Change default jsbundle file name for android

geof90:
To "index.android.bundle", to match RN's default https://github.com/facebook/react-native/blob/master/local-cli/generator-android/templates/src/app/react.gradle#L5

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -288,6 +288,7 @@ After configuring your React Native app to query for updates against the CodePus
 
 ```
 code-push release-react <appName> <platform> 
+[--bundleName <bundleName>]
 [--deploymentName <deploymentName>]
 [--description <description>]
 [--entryFile <entryFile>]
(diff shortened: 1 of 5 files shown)


lostintangent:
Do you think we should go ahead and add a new flag to the `release-react` command that allows specifying the bundle name? That way, this fix ensures the E2E works for the default file names, but if someone changes the name, they can configure it in their Obj-C/Java code, and customize it when releasing.

geof90:
Is `--bundleName` ok?

lostintangent:
Yep, and maybe allow "-b" as an alias?

geof90:
@lostintangent done

lostintangent:
Could you also update the CLI readme?

geof90:
Yep done!

lostintangent:
LGTM!

PR #145, opened 2016-02-29
Login session file changes

silhouettes:
Firstly, reflect the upcoming server change by removing the concept of a base 64-encoded 'access token', which was unnecessary. We consistently use access keys now.

Secondly, ignore the `serverUrl` property in the session file. The reason is that many users will have session files on disk that point to `https://codepush.azurewebsites.net`, whereas in the upcoming release the CLI will communicate with `https://codepush-management.azurewebsites.net`. For these users, the server would erroneously send them a message asking them to upgrade their CLI (even if they're already on the latest!) For debugging purposes, I've instead added a `customServerUrl` property that only gets written to disk if explicitly specified (the URL will be pulled from the SDK otherwise).

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -161,7 +161,7 @@ Once confirmed, the specified developer becomes the app's owner and immediately 
 
 ## Deployment management
 
-From the CodePush perspective, an app is simply a named grouping for one or more things called "deployments". While the app represents a conceptual "namespace" or "scope" for a platform-specific version of an app (e.g. the iOS port of Foo app), it's deployments represent the actual target for releasing updates (for developers) and synchronizing updates (for end-users). Deployments allow you to have multiple "environments" for each app in-flight at any given time, and help model the reality that apps typically move from a dev's personal environment to a testing/QA/staging environment, before finally making their way into production.
+From the CodePush perspective, an app is simply a named grouping for one or more things called "deployments". While the app represents a conceptual "namespace" or "scope" for a platform-specific version of an app (e.g. the iOS port of Foo app), its deployments represent the actual target for releasing updates (for developers) and synchronizing updates (for end-users). Deployments allow you to have multiple "environments" for each app in-flight at any given time, and help model the reality that apps typically move from a dev's personal environment to a testing/QA/staging environment, before finally making their way into production.
 
 *NOTE: As you'll see below, the `release`, `promote` and `rollback` commands require both an app name and a deployment name is order to work, because it is the combination of the two that uniquely identifies a point of distribution (e.g. I want to release an update of my iOS app to my beta testers).*
 
@@ -264,7 +264,7 @@ This provides an optional "change log" for the deployment. The value is simply r
 
 This specifies whether the update should be considered mandatory or not (e.g. it includes a critical security fix). This attribute is simply round tripped to the client, who can then decide if and how they would like to enforce it.
 
-*NOTE: This parameter is simply a "flag", and therefore, its absence indicates that the release is optional, and it's presence indicates that it's mandatory. You can provide a value to it (e.g. `--mandatory true`), however, simply specifying `--mandatory` is sufficient for marking a release as mandatory.*
+*NOTE: This parameter is simply a "flag", and therefore, its absence indicates that the release is optional, and its presence indicates that it's mandatory. You can provide a value to it (e.g. `--mandatory true`), however, simply specifying `--mandatory` is sufficient for marking a release as mandatory.*
 
 The mandatory attribute is unique because the server will dynamically modify it as necessary in order to ensure that the semantics of your releases are maintained for your end-users. For example, imagine that you released the following three updates to your app:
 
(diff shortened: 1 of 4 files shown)


cli/README.md
```diff
@@ -37,23 +37,23 @@ Every command within the CodePush CLI requires authentication, and therefore, be
...
 ```
 code-push access-key ls
 code-push access-key rm <accessKey>
 ```
 
-If you need additional keys, that can be used to authenticate against the CodePush service without needing to give access to your GitHub and/or Microsoft crendentials, you can run the following command to create one (along with a description of what it is for):
+If you need additional keys that can be used to authenticate against the CodePush service without needing to give access to your GitHub and/or Microsoft credentials, you can run the following command to create a persistent one (along with a description of what it is for):
 
 ```
 code-push access-key add "VSO Integration"
```
lostintangent: Do you want to update this string to "VSTS integration" while you're at it? :)

cli/script/command-executor.ts
```diff
@@ -39,16 +39,18 @@ interface NameToCountMap {
...
-    providerName: string;
-    providerUniqueId: string;
-    serverUrl: string;
+    // The 'providerName' property has been obsoleted
+    // The 'providerUniqueId' property has been obsoleted
+    // The 'serverUrl' property has been obsoleted
 }
 
 interface ILoginConnectionInfo {
     accessKey: string;
```
geof90: What do you think about making this be called "accessKeyName" instead, then we can merge the two interfaces, and not have to care whether the saved one old or the new format?

silhouettes: I think it's too late, because if we do that then we have the opposite problem where some people still have configuration files with "accessKey" as the property, which we then need to special case for backwards compatibility :P

cli/README.md
```diff
@@ -37,23 +37,23 @@ Every command within the CodePush CLI requires authentication, and therefore, be
 code-push login
 ```
 
-This will launch a browser, asking you to authenticate with either your GitHub or Microsoft account. This will generate an access token that you need to copy/paste into the CLI (it will prompt you for it). You are now succesfully authenticated and can safely close your browser window.
+This will launch a browser, asking you to authenticate with either your GitHub or Microsoft account. This will generate an access key that you need to copy/paste into the CLI (it will prompt you for it). You are now succesfully authenticated and can safely close your browser window.
 
-When you login from the CLI, your access token (kind of like a cookie) is persisted to disk so that you don't have to login everytime you attempt to access your account. In order to delete this file from your computer, simply run the following command:
+When you login from the CLI, your access key is persisted to disk for the duration of your session so that you don't have to login everytime you attempt to access your account. In order to end your session and delete this access key, simply run the following command:
```
dtivel: every time

cli/README.md
```diff
@@ -344,7 +344,7 @@ The `promote` command will create a new release for the destination deployment,
 
 1. It's quicker, since you don't need to re-assemble the release assets you want to publish or remember the description/app store version that are associated with the source deployment's release.
 
-2. It's less error-prone, since the promote operartion ensures that the exact thing that you already tested in the source deployment (e.g. `Staging`) will become active in the destination deployment (e.g. `Production`). 
+2. It's less error-prone, since the promote operartion ensures that the exact thing that you already tested in the source deployment (e.g. `Staging`) will become active in the destination deployment (e.g. `Production`).
```
dtivel: operation

cli/README.md
```diff
@@ -344,7 +344,7 @@ The `promote` command will create a new release for the destination deployment,
 
 1. It's quicker, since you don't need to re-assemble the release assets you want to publish or remember the description/app store version that are associated with the source deployment's release.
```
dtivel: reassemble

dtivel:
Looks good.

cli/script/command-executor.ts
```diff
@@ -493,13 +503,13 @@ function getTotalActiveFromDeploymentMetrics(metrics: DeploymentMetrics): number
...
 
-function initiateExternalAuthenticationAsync(serverUrl: string, action: string): void {
+function initiateExternalAuthenticationAsync(action: string, serverUrl?: string): void {
     var message: string = `A browser is being launched to authenticate your account. Follow the instructions ` +
         `it displays to complete your ${action === "register" ? "registration" : "login"}.\r\n`;
 
     log(message);
     var hostname: string = os.hostname();
-    var url: string = serverUrl + "/auth/" + action + "?hostname=" + hostname;
+    var url: string = (serverUrl || AccountManager.SERVER_URL) + "/auth/" + action + "?hostname=" + hostname;
```
geof90: I know this was there before, but this would look way better with a template string

geof90:
LGTM!

silhouettes:
I went ahead and copy pasted the whole README into Word and fixed all of the typos I could find, let me know if I missed anything!

cli/script/command-executor.ts
silhouettes: I think the rule should be that you fix whatever you touch, so I've gone ahead and converted the relevant strings, including this one!

PR #148, opened 2016-03-01
Do not remove the access key on logout if we logged in via '--accessKey'

silhouettes:
This is especially important for the HockeyApp integration, as HockeyApp customers will be logging in with their HockeyApp API key, and need it to persist across multiple sessions.

In future it would make sense to store metadata about how the access key was generated on the server, but regardless of whether we do that it makes sense to me to do this particular piece on the client so that we don't need to consult with the server.

### DIFF
--- a/cli/definitions/cli.ts
+++ b/cli/definitions/cli.ts
@@ -113,7 +113,6 @@ export interface ILoginCommand extends ICommand {
 }
 
 export interface ILogoutCommand extends ICommand {
-    isLocal: boolean;
 }
 
 export interface IPromoteCommand extends ICommand {
--- a/cli/definitions/cli.ts
+++ b/cli/definitions/cli.ts
@@ -112,9 +112,6 @@ export interface ILoginCommand extends ICommand {
     accessKey: string;
 }
 
-export interface ILogoutCommand extends ICommand {
-}
-
 export interface IPromoteCommand extends ICommand {
     appName: string;
     sourceDeploymentName: string;
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -450,7 +450,7 @@ export function execute(command: cli.ICommand): Promise<void> {
                     return login(<cli.ILoginCommand>command);
 
                 case cli.CommandType.logout:
-                    return logout(<cli.ILogoutCommand>command);
+                    return logout(command);
 
                 case cli.CommandType.promote:
                     return promote(<cli.IPromoteCommand>command);
@@ -552,7 +552,7 @@ function loginWithExternalAuthentication(action: string, serverUrl?: string): Pr
         });
 }
 
-function logout(command: cli.ILogoutCommand): Promise<void> {
+function logout(command: cli.ICommand): Promise<void> {
     return Q(<void>null)
         .then((): Promise<void> => {
             if (!connectionInfo.preserveAccessKeyOnLogout) {
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -51,6 +51,7 @@ interface ILoginConnectionInfo {
     accessKey: string;
     // The 'serverUrl' property has been obsoleted
     customServerUrl?: string;   // A custom serverUrl for internal debugging purposes
+    preserveAccessKeyOnLogout?: boolean;
 }
 
 interface IPackageFile {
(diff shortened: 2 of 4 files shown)


geof90:
@ryuyu @lostintangent looks like the VSTS plugin should incorporate this change too!

cli/script/command-executor.ts
```diff
@@ -554,17 +557,25 @@ function loginWithExternalAuthentication(action: string, serverUrl?: string): Pr
...
-                        log("Removed access key " + sdk.accessKey + ".");
-                        sdk = null;
+                        log(`Removed access key ${sdk.accessKey}.`);
                     });
+            } else {
+                log("Warning: Your access key is still valid for future sessions. Please explicitly remove it if desired.");
             }
         })
-        .then((): void => deleteConnectionInfoCache(), (): void => deleteConnectionInfoCache());
+        .catch((err: Error) => delayedError = err)
```
geof90: Seems like we should just use `promise.finally()` here

geof90:
LGTM!

cli/definitions/cli.ts
```diff
@@ -113,7 +113,6 @@ export interface ILoginCommand extends ICommand {
 }
 
 export interface ILogoutCommand extends ICommand {
-    isLocal: boolean;
 }
```
dtivel: Just remove the ILogoutCommand interface.

cli/script/command-parser.ts
```diff
@@ -261,11 +261,10 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
         isValidCommandCategory = true;
         isValidCommand = true;
-        yargs.usage(USAGE_PREFIX + " logout [--local]")
+        yargs.usage(USAGE_PREFIX + " logout")
             .demand(/*count*/ 1, /*max*/ 1)  // Require exactly one non-option argument.
-            .example("logout", "Log out and also remove the access key used for the current session")
-            .example("logout --local", "Log out but allow the use of the same access key for future logins")
-            .option("local", { demand: false, description: "Whether to delete the current session's access key on the server", type: "boolean" });
+            .example("logout", "Log out and end your session");
+            // The 'logout --local' option is obsoleted
```
dtivel: I'd prefer not having "x is obsoleted" comments littered throughout the codebase (here and everywhere else).  Commit/release descriptions can summarize this.  Otherwise, there will be additional work at some future point in time to remove these comments.  Comments should exist to clarify intent of existing code when necessary, not to explain what happened to code of yesteryear.

If you feel strongly about communicating that a feature is obsoleted, you could add that back in the console output when --local is passed.  I don't think this is necessary if release notes communicate it instead.

cli/script/command-parser.ts
```diff
@@ -526,10 +525,6 @@ function createCommand(): cli.ICommand {
 
             case "logout":
```
dtivel: Remove this `case`.

dtivel:
Looks good.

cli/script/command-executor.ts
```diff
@@ -554,17 +557,25 @@ function loginWithExternalAuthentication(action: string, serverUrl?: string): Pr
...
-            if (!command.isLocal) {
+            if (!connectionInfo.preserveAccessKeyOnLogout) {
                 return sdk.removeAccessKey(sdk.accessKey)
                     .then((): void => {
-                        log("Removed access key " + sdk.accessKey + ".");
-                        sdk = null;
+                        log(`Removed access key ${sdk.accessKey}.`);
                     });
+            } else {
+                log("Warning: Your access key is still valid for future sessions. Please explicitly remove it if desired.");
```
lostintangent: Is this warning actually necessary? I feel like the new behavior is intuitive, and therefore, I'm not sure if it makes sense to warn the user here.

lostintangent:
LGTM!

cli/script/command-executor.ts
silhouettes: Nice! This is much cleaner.

cli/definitions/cli.ts
silhouettes: I feel like this breaks consistency and type annotation, e.g. this interface is referred to in lines 457 and 559 of `command-executor.ts`. What do you think?

cli/script/command-parser.ts
silhouettes: I don't think I can remove this case, as we still need to construct the command with the right enum value here.

dtivel: Yup, you're right.

cli/definitions/cli.ts
dtivel: ICommand has the "type" field.  That's all you really need to call the logout(...) function.  The ILogoutCommand interface is otherwise useless.

With your change the cast on line 457 is no longer necessary, and the `command` parameter on line 559 isn't even used.

PR #149, opened 2016-03-03
Let updates target a range of binary versions

geof90:
Allow updates to target a range of binary versions using semver range expressions https://github.com/npm/node-semver#advanced-range-syntax. Added an extra option in the `release-react` command to allow users to override the target binary range value specified in the app metadata. Added some tests for the new changes.

Updated the docs to reflect this new behaviour.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -344,9 +344,9 @@ This is the same parameter as the one described in the [above section](#mandator
 
 This specifies the relative path to where the sourcemap file for resulting update's JS bundle should be generated. If left unspecified, sourcemaps will not be generated.
 
-### Target binary range parameter
+### Target binary version parameter
 
-This is the same parameter as the one described in the [above section](#target-binary-range-parameter). If left unspecified, the command defaults to targeting only the specified version in the project's metadata (`Info.plist` if this update is for iOS clients, and `build.gradle` for Android clients).
+This is the same parameter as the one described in the [above section](#target-binary-version-parameter). If left unspecified, the command defaults to targeting only the specified version in the project's metadata (`Info.plist` if this update is for iOS clients, and `build.gradle` for Android clients).
 
 ## Promoting updates across deployments
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -206,12 +206,12 @@ When the metrics cell reports `No installs recorded`, that indicates that the se
 
 ## Releasing app updates
 
-*NOTE: If your app is built using React Native, we have a different command that automates generating the update contents and inferring some of the parameters (e.g. `targetBinaryRange`) from the project's metadata. Check out the section: [Releasing updates to a React Native app](#releasing-updates-to-a-react-native-app).*
+*NOTE: If your app is built using React Native, we have a different command that automates generating the update contents and inferring some of the parameters (e.g. `targetBinaryVersion`) from the project's metadata. Check out the section: [Releasing updates to a React Native app](#releasing-updates-to-a-react-native-app).*
 
 Once your app has been configured to query for updates against the CodePush service--using your desired deployment--you can begin pushing updates to it using the following command:
 
 ```
-code-push release <appName> <updateContents> <targetBinaryRange>
+code-push release <appName> <updateContents> <targetBinaryVersion>
 [--deploymentName <deploymentName>]
 [--description <description>]
 [--mandatory]
@@ -232,9 +232,15 @@ It's important that the path you specify refers to the platform-specific, prepar
 | React Native wo/assets (iOS)     | `react-native bundle --platform ios --entry-file <entryFile> --bundle-output <bundleOutput> --dev false`                                                   | Value of the `--bundle-output` option                                                                 |
 | React Native w/assets (iOS)      | `react-native bundle --platform ios --entry-file <entryFile> --bundle-output <releaseFolder>/<bundleOutput> --assets-dest <releaseFolder> --dev false` | Value of the `--assets-dest` option, which should represent a newly created directory that includes your assets and JS bundle |
 
-### Target binary range parameter
+### Target binary version parameter
 
-This specifies a [semver range expression](https://github.com/npm/node-semver#advanced-range-syntax) that covers all the store/binary versions that this update applies to. Any client device running a version of the binary that satisfies the range expression (i.e. `semver.satisfies(version, range)` returns `true`) will get the update. Examples of valid semver range expressions are as follows:
+This specifies the store/binary version of the application you are releasing the update for, so that only users running that version will receive the update, while users running an older and/or newer version of the app binary will not. This is useful for the following reasons:
+
+1. If a user is running an older binary version, it's possible that there are breaking changes in the CodePush update that wouldn't be compatible with what they're running.
+
+2. If a user is running a newer binary version, then it's presumed that what they are running is newer (and potentially incompatible) with the CodePush update.
+
+If you ever want an update to target multiple versions of the app store binary, we also allow you to specify the parameter as a [semver range expression](https://github.com/npm/node-semver#advanced-range-syntax). That way, any client device running a version of the binary that satisfies the range expression (i.e. `semver.satisfies(version, range)` returns `true`) will get the update. Examples of valid semver range expressions are as follows:
 
 | Range Expression | Who gets the update                                                                    |
 |------------------|----------------------------------------------------------------------------------------|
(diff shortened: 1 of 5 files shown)


cli/README.md
```diff
@@ -331,6 +341,10 @@ This is the same parameter as the one described in the [above section](#mandator
 
 This specifies the relative path to where the sourcemap file for resulting update's JS bundle should be generated. If left unspecified, sourcemaps will not be generated.
 
+### Target binary range parameter
```
lostintangent: Should we mention again here what the default value is and explain that setting this parameter is valuable if you want to override that?

cli/script/command-executor.ts
```diff
@@ -965,8 +972,16 @@ export var releaseReact = (command: cli.IReleaseReactCommand): Promise<void> =>
             throw new Error(`Entry file "${entryFile}" does not exist.`);
         }
     }
+    
+    if (command.appStoreVersion && semver.validRange(command.appStoreVersion) === null) {
+        throw new Error("Please use a semver-compliant target binary version range, for example \"^1.0.3\".");
```
lostintangent: Maybe add another semver range example? Something like '...for example: "1.0.0", "*" or "^1.2.3")'?

cli/README.md
```diff
@@ -232,15 +232,25 @@ It's important that the path you specify refers to the platform-specific, prepar
...
+This specifies a [semver range expression](https://github.com/npm/node-semver#advanced-range-syntax) that covers all the store/binary versions that this update applies to. Any client device running a version of the binary that satisfies the range expression (i.e. `semver.satisfies(version, range)` returns `true`) will get the update. Examples of valid semver range expressions are as follows:
+
+| Range Expression | Who gets the update                                                                    |
+|------------------|----------------------------------------------------------------------------------------|
+| `1.2.3`          | Only devices running the specific binary app store version `1.2.3` of your app         |
+| `1.2.*`          | Devices running major version 1, minor version 2 and any patch version of your app     |
+| `>=1.2.3 <1.2.7` | Devices running any binary version between `1.2.3` (inclusive) and `1.2.7` (exclusive) |
+| `^1.2.3`         | Equivalent to `>=1.2.3 <2.0.0`                                                         |
+| `*`              | Any device configured to consume updates from your CodePush app                        |
```
lostintangent: For completeness, should we add a tilde range example (e.g. "~1.2.3") and a hyphenated range example (e.g. "1.2.3 - 1.4.0")?

lostintangent:
LGTM!

cli/README.md
```diff
@@ -232,15 +232,30 @@ It's important that the path you specify refers to the platform-specific, prepar
...
 
-This specifies the semver-compliant (e.g. `1.0.0` not `1.0`) store/binary version of the application you are releasing the update for. Only users running this **exact version** will receive the update. Users running an older and/or newer version of the app binary will not receive this update, for the following reasons:
+This specifies a [semver range expression](https://github.com/npm/node-semver#advanced-range-syntax) that covers all the store/binary versions that this update applies to. Any client device running a version of the binary that satisfies the range expression (i.e. `semver.satisfies(version, range)` returns `true`) will get the update. Examples of valid semver range expressions are as follows:
+
+| Range Expression | Who gets the update                                                                    |
+|------------------|----------------------------------------------------------------------------------------|
+| `1.2.3`          | Only devices running the specific binary app store version `1.2.3` of your app         |
+| `*`              | Any device configured to consume updates from your CodePush app                        |
```
lostintangent: It turns out that "" is equivalent to "*", in the same way that "1" is equivalent to "1.*.*". Should we include an example that highlights creating a range by omitting a component (e.g. 1.0, 1)?

cli/README.md
```diff
@@ -232,15 +232,30 @@ It's important that the path you specify refers to the platform-specific, prepar
...
+| `1.2.x`          | Devices running major version 1, minor version 2 and any patch version of your app     |
+| `1.2.3 - 1.2.7`  | Devices running any binary version between `1.2.3` (inclusive) and `1.2.7` (inclusive) |
+| `>=1.2.3 <1.2.7` | Devices running any binary version between `1.2.3` (inclusive) and `1.2.7` (exclusive) |
+| `~1.2.3`         | Equivalent to `>=1.2.3 <1.3.0`                                                         |
+| `^1.2.3`         | Equivalent to `>=1.2.3 <2.0.0`                                                         |
```
lostintangent: I'd also like to understand how we behave with pre-release tags: https://github.com/npm/node-semver#prerelease-tags

cli/script/command-executor.ts
```diff
@@ -741,6 +741,7 @@ function getPackageMetricsString(packageObject: PackageWithMetrics): string {
 }
 
 function getReactNativeProjectAppVersion(platform: string, projectName: string): Promise<string> {
+    var missingPatchVersionRegex = /^\d+\.\d+$/;
```
shishirx34: : RegExp ?

cli/script/command-executor.ts
```diff
@@ -965,8 +966,16 @@ export var releaseReact = (command: cli.IReleaseReactCommand): Promise<void> =>
             throw new Error(`Entry file "${entryFile}" does not exist.`);
         }
     }
+    
+    if (command.appStoreVersion && semver.validRange(command.appStoreVersion) === null) {
+        throw new Error("Please use a semver-compliant target binary version range, for example \"1.0.0\", \"*\" or \"^1.2.3\".");
+    }
```
shishirx34: I see this being used in multiple places, may be refactor the above three lines into a separate method `validateAppVersion` which throws error on invalid range.

cli/script/command-executor.ts
```diff
@@ -840,8 +841,8 @@ function promote(command: cli.IPromoteCommand): Promise<void> {
 export var release = (command: cli.IReleaseCommand): Promise<void> => {
     if (isBinaryOrZip(command.package)) {
         throw new Error("It is unnecessary to package releases in a .zip or binary file. Please specify the direct path to the update content's directory (e.g. /platforms/ios/www) or file (e.g. main.jsbundle).");
-    } else if (semver.valid(command.appStoreVersion) === null) {
-        throw new Error("Please use a semver-compliant app store version, for example \"1.0.3\".");
+    } else if (semver.validRange(command.appStoreVersion) === null) {
+        throw new Error("Please use a semver-compliant target binary version range, for example \"1.0.0\", \"*\" or \"^1.2.3\".");
```
silhouettes: Do we need to do this validation on the client-side if we have server-side validation already?

shishirx34:
LGTM

silhouettes:
Looks good

cli/script/command-executor.ts
geof90: I think this error message is friendlier? The server, using our validation library, will just say something like "[Error] appVersion is invalid".

shishirx34: I think we should still do some client side validation to avoid unnecessary server hits for the errors we know are going to be thrown.

silhouettes: I don't think we need to optimize the performance in error cases like this. More importantly, I think client-side validation comes with a maintenance cost, as well as the cost of having less flexibility to change things on the server (which we've already run into a few times).

I agree with having a friendlier error message, and ideally that would be generated on the server, but I understand if it's too much work to do it as part of this PR.

cli/README.md
lostintangent: Rename this to `Target binary version`?

cli/script/command-parser.ts
```diff
@@ -306,7 +306,8 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
             .option("mandatory", { alias: "m", default: false, demand: false, description: "Whether this update should be considered mandatory to the client", type: "boolean" })
-            .option("sourcemapOutput", { alias: "s", default: null, demand: false, description: "The path to where the sourcemap for the resulting bundle should be stored. If unspecified, sourcemaps will not be generated.", type: "string" });
+            .option("sourcemapOutput", { alias: "s", default: null, demand: false, description: "The path to where the sourcemap for the resulting bundle should be stored. If unspecified, sourcemaps will not be generated.", type: "string" })
+            .option("targetBinaryRange", { alias: "t", default: null, demand: false, description: "The semver range expression spanning all the binary app store versions that should get this update. If omitted, the update will default to target only the same version as the current binary version specified in \"Info.plist\" (iOS) or \"build.gradle\" (Android)", type: "string" });
```
lostintangent: targetBinaryVersion?

PR #150, opened 2016-03-04
Productize the management SDK

silhouettes:
Some quick finishing touches to the SDK for public consumption:

1. Make the management SDK the default export
1. Rename methods to be consistent with CLI commands (and remove the use of the term 'package')
       - I've also renamed the `/deployments/<name>/packageHistory` route to `/deployments/<name>/history` (for which I'll make a quick server-side change), because having both `packageHistory` and `deploymentHistory` as names for the same thing is kind of confusing.
1. Add a hook for custom headers so that the CLI can send the `X-CodePush-CLI-Version` header that we need for app insights logging


sdk/script/management-sdk.ts
```diff
@@ -208,7 +195,17 @@ export class AccountManager {
             .then(() => null);
     }
 
-    public releasePackage(appName: string, deploymentName: string, fileOrPath: File | string, description: string, appVersion: string, isMandatory: boolean = false, uploadProgressCallback?: (progress: number) => void): Promise<void> {
+    public getDeploymentMetrics(appName: string, deploymentName: string): Promise<DeploymentMetrics> {
```
lostintangent: Should we expose this? We never exposed this via the CLI, so I wonder if we need to bother making it available in the SDK. It's probably fine as is.

sdk/script/management-sdk.ts
```diff
@@ -208,7 +195,17 @@ export class AccountManager {
...
+        return this.get(urlEncode `/apps/${appName}/deployments/${deploymentName}/metrics`)
+            .then((res: JsonResponse) => res.body.metrics);
+    }
+
+    public getDeploymentHistory(appName: string, deploymentName: string): Promise<Package[]> {
+        return this.get(urlEncode `/apps/${appName}/deployments/${deploymentName}/history`)
+            .then((res: JsonResponse) => res.body.history);
+    }
+
+    public release(appName: string, deploymentName: string, fileOrPath: File | string, description: string, appVersion: string, isMandatory: boolean = false, uploadProgressCallback?: (progress: number) => void): Promise<void> {
```
lostintangent: We should change the "appVersion" parameter to be called "targetBinaryVersion" for consistency with the CLI.

sdk/script/management-sdk.ts
```diff
@@ -193,11 +185,6 @@ export class AccountManager {
             .then((res: JsonResponse) => res.body.deployment);
     }
 
-    public getDeploymentMetrics(appName: string, deploymentName: string): Promise<DeploymentMetrics> {
-        return this.get(urlEncode `/apps/${appName}/deployments/${deploymentName}/metrics`)
-            .then((res: JsonResponse) => res.body.metrics);
-    }
-
     public updateDeployment(appName: string, deploymentName: string, infoToChange: Deployment): Promise<void> {
```
lostintangent: Should we expose "renameDeployment" and "renameApp" methods to better reflect the CLI and the general purpose of editing either entity type? If this method is needed by the CLI, maybe we could just not document it, and create a rename variant to simply "wraps" the more general-purpose update methods.

sdk/script/management-sdk.ts
```diff
@@ -58,18 +44,24 @@ function urlEncode(strings: string[], ...values: string[]): string {
     return result;
 }
 
-export class AccountManager {
+class AccountManager {
+    public static PERMISSIONS = {
```
lostintangent: Do we want this name to be uppercased? I had thought we were naming enums using sentence-casing, with their members being uppercased (e.g. InstallMode.IMMEDIATE).

lostintangent: I think I'd be in favor of calling it "AppPermission" as well, unless we're not thinking of this as an "enum".

sdk/script/management-sdk.ts
```diff
@@ -58,18 +44,24 @@ function urlEncode(strings: string[], ...values: string[]): string {
...
-    private _userAgent: string;
+    private _customHeaders: Headers;
+
+    constructor(accessKey: string, customHeaders?: Headers, serverUrl?: string) {
+        if (!accessKey) throw new Error("An access key must be specified.");
 
-    constructor(accessKey: string, userAgent?: string, serverUrl?: string) {
         this._accessKey = accessKey;
-        this._userAgent = userAgent;
+        this._customHeaders = customHeaders;
```
geof90: Would there be a value in also having a default header for the SDK, and then merge the supplied headers with ours?

sdk/script/management-sdk.ts
geof90: Should we move "description" to the right of "targetBinaryVersion" and make it optional?

sdk/script/management-sdk.ts
dtivel: +1 for sentence-casing Permissions and keeping members upper-cased.

dtivel:
Looks good.

sdk/script/management-sdk.ts
silhouettes: Yeah I think so, I'll go ahead and change that

lostintangent:
LGTM!

sdk/script/management-sdk.ts
```diff
@@ -208,9 +190,19 @@ export class AccountManager {
...
+        return this.get(urlEncode `/apps/${appName}/deployments/${deploymentName}/metrics`)
+            .then((res: JsonResponse) => res.body.metrics);
+    }
+
+    public getDeploymentHistory(appName: string, deploymentName: string): Promise<Package[]> {
+        return this.get(urlEncode `/apps/${appName}/deployments/${deploymentName}/history`)
+            .then((res: JsonResponse) => res.body.history);
+    }
+
+    public release(appName: string, deploymentName: string, fileOrPath: File | string, targetBinaryVersion: string, description: string, isMandatory: boolean = false, uploadProgressCallback?: (progress: number) => void): Promise<void> {
```
geof90: `description?` with the question mark

geof90:
LGTM!

PR #152, opened 2016-03-07
Expose --dev option to release-react

geof90:
Just a super quick, 5min change that exposes the --dev option to the release-react command since I figured it would be useful for my demo.


lostintangent:
LGTM! Do you want to update the readme as well? BTW, for my ReactConf demo, I had change the dev option to be true in my local CLI install precisely for the reason you wanted to make this change :)

geof90:
Updated the docs, merging.

PR #153, opened 2016-03-07
Version bump

lostintangent:
This simply bumps the management SDK and CLI versions in preparation for our upcoming release. The current `code-push` dependency that CLI has is incorrect, since it would fail to work correctly when using the `1.6.0-beta` release.


silhouettes:
Looks good to me!!

PR #154, opened 2016-03-08
Clear updates associated with a deployment

geof90:
E.g.
```
$ code-push deployment clear MyApp Staging
```

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -434,4 +434,4 @@ You can clear the release history associated with a deployment using the followi
 code-push deployment history <appName> <deploymentName>
 ```
 
-After running this command, client devices configured to receive updates from this deployment using its associated deployment key will no longer be able to receive the updates that have been cleared. This command is irreversible, and therefore should not be used in a production deployment.
+After running this command, client devices configured to receive updates from this deployment using its associated deployment key will no longer receive those updates that have been cleared. This command is irreversible, and therefore should not be used in a production deployment.
--- a/cli/README.md
+++ b/cli/README.md
@@ -431,7 +431,7 @@ By default, the history doesn't display the author of each release, but if you a
 You can clear the release history associated with a deployment using the following command: 
 
 ```
-code-push deployment history <appName> <deploymentName>
+code-push deployment clear <appName> <deploymentName>
 ```
 
 After running this command, client devices configured to receive updates from this deployment using its associated deployment key will no longer receive those updates that have been cleared. This command is irreversible, and therefore should not be used in a production deployment.
--- a/cli/README.md
+++ b/cli/README.md
@@ -172,8 +172,8 @@ code-push deployment add <appName> <deploymentName>
 Just like with apps, you can remove and rename deployments as well, using the following commands respectively:
 
 ```
-code-push deployment rename <appName> <deploymentName> <newDeploymentName>
 code-push deployment rm <appName> <deploymentName>
+code-push deployment rename <appName> <deploymentName> <newDeploymentName>
 ```
 
 If at any time you'd like to view the list of deployments that a specific app includes, you can simply run the following command:
@@ -425,3 +425,13 @@ Additionally, the history displays the install metrics for each release. You can
 By default, the history doesn't display the author of each release, but if you are collaborating on an app with other developers, and want to view who released each update, you can pass the additional `--displayAuthor` (or `-a`) flag to the history command.
 
 *NOTE: The history command can also be run using the "h" alias*
+
+## Clearing release history
+
+You can clear the release history associated with a deployment using the following command: 
+
+```
+code-push deployment history <appName> <deploymentName>
+```
+
+After running this command, client devices configured to receive updates from this deployment using its associated deployment key will no longer be able to receive the updates that have been cleared. This command is irreversible, and therefore should not be used in a production deployment.
--- a/cli/definitions/cli.ts
+++ b/cli/definitions/cli.ts
@@ -11,6 +11,7 @@
     collaboratorList,
     collaboratorRemove,
     deploymentAdd,
+    deploymentClear,
     deploymentHistory,
     deploymentList,
     deploymentMetrics,
@@ -79,6 +80,11 @@ export interface ICollaboratorRemoveCommand extends ICommand {
 }
 
 export interface IDeploymentAddCommand extends ICommand {
+    appName: string;
+    deploymentName: string;
+}
+
+export interface IDeploymentClearCommand extends ICommand {
     appName: string;
     deploymentName: string;
 }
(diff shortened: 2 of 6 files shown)


cli/script/command-parser.ts
```diff
@@ -113,6 +113,15 @@ function removeCollaborator(commandName: string, yargs: yargs.Argv): void {
     addCommonConfiguration(yargs);
 }
 
+function deploymentClear(commandName: string, yargs: yargs.Argv): void {
+    isValidCommand = true;
+    yargs.usage(USAGE_PREFIX + " deployment " + commandName + " <appName> <deploymentName>")
+        .demand(/*count*/ 4, /*max*/ 4)  // Require exactly four non-option arguments.
+        .example("deployment " + commandName + " MyApp MyDeployment", "Clears away all the updates associated with deployment \"MyDeployment\" from app \"MyApp\"");
```
dtivel: Clears ~~away~~ all ~~the~~ updates associated...

cli/script/command-parser.ts
```diff
@@ -227,6 +236,7 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
 
                 addCommonConfiguration(yargs);
             })
+            .command("clear", "Clears away all the updates associated with a deployment", (yargs: yargs.Argv) => deploymentClear("clear", yargs))
```
dtivel: Clears ~~away~~ all ~~the~~ updates associated

cli/script/command-executor.ts
```diff
@@ -275,6 +275,20 @@ function deploymentAdd(command: cli.IDeploymentAddCommand): Promise<void> {
         });
 }
 
+function deploymentClear(command: cli.IDeploymentAddCommand): Promise<void> {
+    return confirm()
+        .then((wasConfirmed: boolean): Promise<void> => {
+            if (wasConfirmed) {
+                return sdk.clearDeployment(command.appName, command.deploymentName)
+                    .then((): void => {
+                        log("Successfully cleared all the updates associated with the \"" + command.deploymentName + "\" deployment from the \"" + command.appName + "\" app.");
```
dtivel: ...cleared all ~~the~~ updates...

...sounds better.

sdk/script/management-sdk.ts
```diff
@@ -169,6 +169,11 @@ class AccountManager {
         return this.post(urlEncode `/apps/${appName}/deployments/`, JSON.stringify(deployment), /*expectResponseBody=*/ true)
             .then((res: JsonResponse) => res.body.deployment);
     }
+    
+    public clearDeployment(appName: string, deploymentName: string): Promise<void> {
```
dtivel: This would be better off named `clearDeploymentHistory`.

cli/definitions/cli.ts
```diff
@@ -83,6 +84,11 @@ export interface IDeploymentAddCommand extends ICommand {
     deploymentName: string;
 }
 
+export interface IDeploymentClearCommand extends ICommand {
```
dtivel: In general the commands should indicate that deployment history is being cleared, not the deployment itself (e.g.:  IDeploymentHistoryClearCommand)

sdk/script/management-sdk.ts
lostintangent: Aren't those two names basically semantically equivalent? The deployment is essentially just a history of releases, so I believe simply calling it clear is clear enough. I don't feel super strongly about the SDK name, but I definitely am in favor of keeping the CLI command "code-push deployment clear" as opposed to "code-push deployment clear-history" or something like that.

cli/script/command-parser.ts
lostintangent: I'd vote for "Clears the release history associated with a deployment.", that way, we ensure the details about what it is doing is clear without needing to encode it into the command name (e.g. "clear-history").

cli/script/command-executor.ts
lostintangent: How about "Successfully cleared the release history associated with..."?

sdk/script/management-sdk.ts
dtivel: No, I don't think they are semantically equivalent.

A CodePush deployment is an update channel (or "environment").  It has its own descriptors and relationships.

As of this change, deployment history is an erasable log of updates released to that channel.

A deployment can have zero to many associated updates.  In other words, a deployment can have deployment history or no deployment history at all, such as after clearing deployment history.  However, clearing deployment history does not change the deployment itself (its existence, its name, its deployment key, etc.).  The channel is still there and usable but waiting for a new release.

This feature erases deployment history.

>>  The deployment is essentially just a history of releases

I think that's a bit of an oversimplification, and at a high-level you may be right.

lostintangent: Besides a deployment's name, key, and its relationship to its parent app, what other attributes does it have other than its history? I think of the clear operation as being equivalent to clearing an array or dictionary/etc., which wouldn't obviously impact the references that point at that dictionary instance (the app->deployment relationship), or even the "identity" (e.g. memory address of the start of the data structure). The name and key are the identity of the deployment (on the management and acquisition sides, respectively), and I guess I'm just saying that I don't think folks would expect an operation named "clear" to mutate those. But I could obviously be over wrong.

I'm definitely not trying to oversimplify this, but I also don't want to overcomplicate it either. Conceptually thinking of a deployment as a history of releases seems legitimate to me.

geof90: I think I will rename all the code methods to be renamed to `clearDeploymentHistory` because it also has parity with the server endpoint being a DELETE on the `/history` route.

And then I will leave the CLI command as `deployment clear`. I think it is just shorter than `deployment clear-history`, and with the CLI helper instructions being printed out, I don't think anyone will be confused.

 Is this an okay compromise?

dtivel: Sounds fine.  Thanks.

geof90:
I've also taken a stab at updating the docs

cli/README.md
```diff
@@ -425,3 +425,13 @@ Additionally, the history displays the install metrics for each release. You can
 By default, the history doesn't display the author of each release, but if you are collaborating on an app with other developers, and want to view who released each update, you can pass the additional `--displayAuthor` (or `-a`) flag to the history command.
 
 *NOTE: The history command can also be run using the "h" alias*
+
+## Clearing release history
+
+You can clear the release history associated with a deployment using the following command: 
+
+```
+code-push deployment history <appName> <deploymentName>
```
shishirx34: don't you mean `code-push deployment clear <appName> <deploymentName>` ?

shishirx34:
LGTM

dtivel:
Looks good.

PR #155, opened 2016-03-08
Change the SDK's addAccessKey() method to be more consumable

silhouettes:
Infer the hostname internally within the method, and make the description field compulsory (in TypeScript at least).

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -96,8 +96,7 @@ export var confirm = (): Promise<boolean> => {
 }
 
 function accessKeyAdd(command: cli.IAccessKeyAddCommand): Promise<void> {
-    var hostname: string = os.hostname();
-    return sdk.addAccessKey(hostname, command.description)
+    return sdk.addAccessKey(command.description)
         .then((accessKey: AccessKey) => {
             log("Successfully created a new access key" + (command.description ? (" \"" + command.description + "\"") : "") + ": " + accessKey.name);
         });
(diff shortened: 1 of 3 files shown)


lostintangent:
LGTM!

geof90:
Actually i think it would make more sense for the hostname to be customizable (and possibly renamed to something like "createdFrom"). We default it to the host name for the CLI, but there may be cases we want to explicitly annotate it with something else e.g. created from HockeyApp servers. What do you think?

geof90:
Also if we wanted the description to be mandatory, we could throw an exception if it is not provided.

But in terms of doing what it's intended to do, this change LGTM.

lostintangent:
I don't think the concept of a machine name makes sense with an explicitly created access key. The description is meant to encode the purpose of the key. The machine name IMO is only valuable for login-generated keys, since they indicate where the session was started, and help indicate if you want to remove it remotely. The current way that we visualize access keys in the CLI is non-ideal, and I would prefer that we add a `Type` column that is either `Login` or `Custom` (or whatever), and the `Description` column would display the machine name for login keys and the description for custom keys.

lostintangent:
When you create an access token or SSH key on GitHub, you specify a description, and they don't capture the machine name it was created from.

silhouettes:
Yeah, I think of the machine name more like an internal or 'generated' field that users don't specify directly, similarly to the `createdTime` field. The only reason it's on the client as opposed to the server (unlike `createdTime`) is that the server can't infer your machine name otherwise. If the user wants to annotate specific information, they can always use the `description` field. Let me know if you disagree!

PR #160, opened 2016-03-11
Gradle parsing issue

geof90:
The gradle-to-js npm module will parse a versionName of "1.0" as the Javascript number "1". With that, we have no way to tell whether it has a minor version or not. I forked the project and changed the parser to always return strings as the obvious but workable solution for our purposes. https://github.com/geof90/gradle-to-js/commit/ccd3362ad227203319f799dda03a6db3577c56a8


geof90:
I've also verified that the plist library (for iOS) doesn't run into the same issue.

lostintangent:
LGTM!

PR #162, opened 2016-03-13
Deleting release zip on erorr

### LINKED ISSUE #161: [CLI] Is it possible to tell code-push cli to create archives in a separate directory?
[CLI] Is it possible to tell code-push cli to create archives in a separate directory? The problem is that right now cli creates zip files in the root of the project over time this becomes really messy

lostintangent:
This resolves #161 by ensuring that the ZIP generated by the `release` command is always deleted, regardless if the release fails (e.g. because the server rejected an identical release).

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -926,17 +926,28 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
     var uploadProgress = (currentProgress: number): void => {
         progressBar.tick(currentProgress - lastTotalProgress);
         lastTotalProgress = currentProgress;
-    }
+    };
 
+    var deleteFileIfTemporary = (file: IPackageFile): void => {
+        if (file.isTemporary) {
+            fs.unlinkSync(filePath);
+        }
+    };
+    
     return getPackageFilePromise
         .then((file: IPackageFile): Promise<void> => {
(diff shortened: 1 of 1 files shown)


cli/script/command-executor.ts
```diff
@@ -926,17 +926,28 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
...
                 .then((): void => {
                     log("Successfully released an update containing the \"" + command.package + "\" " + (isSingleFilePackage ? "file" : "directory") + " to the \"" + command.deploymentName + "\" deployment of the \"" + command.appName + "\" app.");
 
-                    if (file.isTemporary) {
-                        fs.unlinkSync(filePath);
-                    }
+                    deleteFileIfTemporary(file);
+                })
+                .catch((error): void => {
+                    deleteFileIfTemporary(file);
```
geof90: I would use .finally() instead of calling deleteFileIfTemporary() in two places

lostintangent: I was going to, but I decided to mirror what you did in the `release-react` command when deleting the temp folder. Since `Promise.prototype.finally` isn't standard, I kind of prefer doing this, but I'm happy either way.

geof90:
LGTM!

PR #176, opened 2016-03-21
Add "disabled" metadata for promote and release + forward shouldRunBinaryVersion flag from server

geof90:
This change allows users to specify the "disabled" metadata when releasing or promoting an update.

Also combined this PR with https://github.com/Microsoft/code-push/pull/174, which configures the acquisition SDK to forward the `shouldRunBinaryVersion` flag sent by the server to notify client plugins to trigger a rollback to the binary version if needed.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -306,7 +306,7 @@ This provides an optional "change log" for the deployment. The value is simply r
 
 #### Disabled parameter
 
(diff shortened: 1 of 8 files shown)


lostintangent:
LGTM!

geof90:
@lostintangent I updated the docs.

PR #181, opened 2016-03-23
Compensate for git bash on Windows when parsing custom server urls

silhouettes:
This is a quick tweak purely for the sake of internal testing, as custom server URL's are not needed by end users.

A major annoyance for the longest time when using Git Bash on Windows is that, due to path translation, a URL like `http://localhost:3000` passed on the command-line ends up as `http:\localhost:3000` by the time it ends up in our CLI. This just compensates for that by translating it back.

### DIFF
--- a/cli/script/command-parser.ts
+++ b/cli/script/command-parser.ts
@@ -729,7 +729,7 @@ function getServerUrl(url: string): string {
         url = url.substring(0, url.length - 1);
     }
 
-    url = url.replace(/^(https?):\\/, "$1://");     // Replace 'http:\' with 'http://' for Windows Git Bash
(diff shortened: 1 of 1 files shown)


cli/script/command-parser.ts
```diff
@@ -729,6 +729,8 @@ function getServerUrl(url: string): string {
         url = url.substring(0, url.length - 1);
     }
 
+    url = url.replace(/^(https?):\\/, "$1://");     // Replace 'http:\' with 'http://' for Windows Git Bash
```
shishirx34: // Replace 'http**(s)**:\' with 'http**(s)**://' for Windows Git Bash

shishirx34:
LGTM

PR #182, opened 2016-03-23
switched process call to exec instead of spawn

ryuyu:
This addresses and issue with release-cordova on Windows.

spawn can't execute bat/cmd files, which is how npm packages are executed on Windows.
Exec doesn't care, so swap to using this.

@shishirx34 @silhouettes @geof90 @lostintangent


lostintangent:
LGTM!

geof90:
LGTM!

PR #199, opened 2016-04-04
Update rest definitions for improving diffs

shishirx34:
Updating the SDK version to reflect change of object structure for package.

Although none of the clients use this change, keeping the definitions in sync with the server change.


geof90:
LGTM!

silhouettes:
Looks good.

PR #201, opened 2016-04-11
Add `build` option to `release-cordova` command

dtivel:
Fixes https://github.com/Microsoft/cordova-plugin-code-push/issues/92.

### DIFF
--- a/cli/definitions/cli.ts
+++ b/cli/definitions/cli.ts
@@ -162,6 +162,7 @@ export interface IReleaseCommand extends IReleaseBaseCommand {
 }
 
 export interface IReleaseCordovaCommand extends IReleaseBaseCommand {
+    build: boolean;
     platform: string;
 }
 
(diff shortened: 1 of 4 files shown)


lostintangent:
I'm not sure if we want to always run `cordova build`, since it would presumably be quite a bit slower than the `prepare` since it has to actually compile the native binary. It's definitely much simpler this way than to add a new flag to switch from using `prepare` to `build` (e.g. `code-push release-cordova MyApp ios --build`), but I'm still on the fence about which solution to this I prefer.

scottbommarito:
I'd rather have a flag. I don't know how many people use Ionic compared to how many people use plain cordova, but slowing down everybody just for a couple people to be able to use the command is not ideal.

candidodmv:
I accept that was said by @lostintangent, this change will affect everyone, but we need to consider that the most part of people are using the approach "prepare"

lostintangent:
LGTM!

PR #208, opened 2016-04-16
add system-wide proxy support

kiliwalk:


sdk/script/management-sdk.ts
```diff
@@ -8,6 +8,10 @@ import Promise = Q.Promise;
 
 import { AccessKey, Account, App, CodePushError, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, UpdateMetrics } from "./types";
 
+var superproxy = require("superagent-proxy");
+superproxy(superagent);
+var proxy = process.env.http_proxy;
```
silhouettes: Thanks so much for contributing this!! Could we make this environment variable upper-case (`HTTP_PROXY`) to match our internal convention?

sdk/script/management-sdk.ts
```diff
@@ -67,7 +71,7 @@ class AccountManager {
 
     public isAuthenticated(): Promise<boolean> {
         return Promise<boolean>((resolve, reject, notify) => {
-            var request: superagent.Request<any> = superagent.get(this._serverUrl + urlEncode `/authenticated`);
+            var request: superagent.Request<any> = (<any>superagent.get(this._serverUrl + urlEncode `/authenticated`)).proxy(proxy);
```
silhouettes: Just to make sure we run cleanly if the proxy is not set, could we change these three instances to have this on another line:

`if (proxy) (<any>request).proxy(proxy);`

silhouettes:
Thank you @kiliwalk! I just have two quick tweaks I'd recommend, and if someone else can sign off on this I'll merge it asap.

kiliwalk:
thank your advice @silhouettes .
last day i wanted to add superagent-proxy typescript definition, but i was failed, so i had to use type-cast to by-pass the type check to call the `proxy` method:D

sdk/script/management-sdk.ts
```diff
@@ -68,6 +72,7 @@ class AccountManager {
     public isAuthenticated(): Promise<boolean> {
         return Promise<boolean>((resolve, reject, notify) => {
             var request: superagent.Request<any> = superagent.get(this._serverUrl + urlEncode `/authenticated`);
+            if(proxy) (<any>request).proxy(proxy);
```
geof90: Nit: if<space>(proxy) ...

sdk/script/management-sdk.ts
```diff
@@ -8,6 +8,10 @@ import Promise = Q.Promise;
 
 import { AccessKey, Account, App, CodePushError, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, UpdateMetrics } from "./types";
 
+var superproxy = require("superagent-proxy");
+superproxy(superagent);
+var proxy = process.env.HTTPS_PROXY || process.env.https_proxy || process.env.HTTP_PROXY || process.env.http_proxy;
```
geof90: I prefer not to have the Management SDK read environment variables directly, because it is designed to be used as a general library. I prefer for the proxy behaviour be explicitly determined by the application using it.

Would this behaviour be more transparent if the `proxy` setting is passed to in the constructor? We can have the CLI read the environment variable and pass the proxy setting to it instead. Either that, or have a `code-push login --proxy <proxyUrl>` option, which saves the proxy setting in the login credentials. What do you think, @silhouettes

geof90:
Some minor comments, LGTM otherwise!

kiliwalk:
@geof90 agreed with the environment vars handled. as to the `proxy` option, maybe we need a global setting across the login users?

kiliwalk:
@geof90 add two option `proxy` and `noProxy` in the login and register command. 
* default the cli command will read the system environment var(like HTTPS_PROXY/HTTP_PROXY). 
* use `--proxy` to specify code-push used-only proxy
* use `--noProxy` to suppress the system environment var when user want to make the code-push independent of the system-wide proxy setting.

lostintangent:
@kilowalk I really like this solution! Thank you so much for taking the time to contribute this :D

cli/script/command-executor.ts
```diff
@@ -1290,4 +1297,10 @@ function whoami(command: cli.ICommand): Promise<void> {
         .then((account): void => {
             log(`${account.email} (${account.linkedProviders.join(", ")})`);
         });
-}
\ No newline at end of file
+}
+
+function getProxy(proxy: string, noProxy?: boolean): string {
```
lostintangent: Isn't the `proxy` parameter also optional?

cli/script/command-parser.ts
```diff
@@ -273,7 +273,11 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
             .demand(/*count*/ 1, /*max*/ 2)  // Require one non-optional and one optional argument.
             .example("login", "Logs in to the CodePush server")
             .example("login --accessKey mykey", "Logs in on behalf of the user who owns and created the access key \"mykey\"")
+            .example("login --proxy proxyURL", "Logs in with the specified proxy url")
```
lostintangent: To help illustrate the fact that this parameter expects a complete URL, can we replace "proxyURL" with an example like "https://someproxy.com:455"?

cli/script/command-parser.ts
```diff
@@ -273,7 +273,11 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
+            .example("login --noProxy", "Logs in without any proxy set in session file or environment vars")
             .option("accessKey", { alias: "key", default: null, demand: false, description: " Access key to authenticate against the CodePush server with, instead of providing your username and password credentials", type: "string" })
+            .option("proxy", { default: null, demand: false, description: " Proxy url", type: "string" })
+            .option("noProxy", { default: false, demand: false, description: " Suppress any proxy set in session file or environment vars", type: "boolean" })
```
lostintangent: How about using this description: "Bypass the system-wide proxy settings"?

lostintangent:
This doesn't need to happen as part of this PR, but it might make sense to also display the proxy server in the output of the "whoami" command, so that users could easily reference if they're current login session is using a proxy.

cli/script/command-parser.ts
```diff
@@ -273,7 +273,11 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
             .demand(/*count*/ 1, /*max*/ 2)  // Require one non-optional and one optional argument.
             .example("login", "Logs in to the CodePush server")
             .example("login --accessKey mykey", "Logs in on behalf of the user who owns and created the access key \"mykey\"")
+            .example("login --proxy proxyURL", "Logs in with the specified proxy url")
+            .example("login --noProxy", "Logs in without any proxy set in session file or environment vars")
```
lostintangent: I would be in favor of omitting the noProxy example to keep things simple. I think showing the examples with and without a proxy server should accommodate the most common cases, and we can just rely on the docs and option description to explain the noProxy parameter.

cli/script/command-parser.ts
```diff
@@ -273,7 +273,11 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
             .option("accessKey", { alias: "key", default: null, demand: false, description: " Access key to authenticate against the CodePush server with, instead of providing your username and password credentials", type: "string" })
+            .option("proxy", { default: null, demand: false, description: " Proxy url", type: "string" })
```
lostintangent: How about using this description: "URL of the proxy server to use"? There's no need to add the leading space either. It's a bug that the access key option has one :)

sdk/script/management-sdk.ts
```diff
@@ -52,13 +55,15 @@ class AccountManager {
     private _accessKey: string;
     private _serverUrl: string;
     private _customHeaders: Headers;
+    private _proxy: string;
 
-    constructor(accessKey: string, customHeaders?: Headers, serverUrl?: string) {
+    constructor(accessKey: string, customHeaders?: Headers, serverUrl?: string, proxy?: string) {
```
lostintangent: I'd be in favor of converting the last three parameters into an options object since they're all optional. That way, this constructor only has two parameters, and a user wouldn't need to specify two nulls simply to specific a proxy server.

cli/script/command-parser.ts
kiliwalk: haha, a bug, you are so serious:D.

cli/script/command-parser.ts
kiliwalk: that's ok, sorry for my poor english. anything like these description and comments, if not much proper, please let me know

lostintangent: Your English is totally fine! We all nit pick each other's docs :)

cli/script/command-executor.ts
kiliwalk: em, i think optional or not are both ok, since currently all places called `getProxy` pass two args. but for future usage, optional is more proper

lostintangent:
LGTM! None of my comments are deal breakers, so I'm cool with whatever you decide. Like I said, this is really awesome to have in the product, so I really appreciate the PR.

kiliwalk:
Hi @lostintangent , you said display the proxy info in `whoami` command. So, if the user have not set any proxy option in login/register, but the system-wide proxy is set, should we display the proxy?

lostintangent:
@kiliwalk If the system-wide setting was set at the time the user logged in or registered, then yeah, we should display that in the whoami, since that proxy would be persisted to their login session file and used until they logout.

kiliwalk:
@lostintangent and shows what message? like 
```
your@email.com (GitHub)
Behind the proxy: http://someproxy.com:445
```

lostintangent:
Maybe just "Proxy:" as the label text? And only display that like if the proxy is non-null in the session file?

kiliwalk:
@lostintangent em, my current solution will not persist the system-wide proxy in the session file, instead if the user login/register with `proxy` set, save the `proxy` option in the session file. Never will save the system-wide proxy, since they may be changed and i want cli to keep sync with that

lostintangent:
@kiliwalk Ah right, that makes sense. I suppose it only makes sense to display the proxy in the whoami output if it was explicitly set in the CLI when logging in.

cli/script/command-executor.ts
```diff
@@ -1289,5 +1296,16 @@ function whoami(command: cli.ICommand): Promise<void> {
     return sdk.getAccountInfo()
         .then((account): void => {
             log(`${account.email} (${account.linkedProviders.join(", ")})`);
+
+            var connectionInfo = deserializeConnectionInfo();
+            if (!connectionInfo.noProxy && connectionInfo.proxy) {
+                log(`Proxy: ${connectionInfo.proxy}`);
```
lostintangent: Should we display something if the `noProxy` flag was set? Otherwise, the user might get confused why CodePush isn't using a system-wide setting. Maybe we could display something like "Proxy: Ignored" to indicate this?

cli/script/command-executor.ts
```diff
@@ -1289,5 +1296,16 @@ function whoami(command: cli.ICommand): Promise<void> {
     return sdk.getAccountInfo()
         .then((account): void => {
             log(`${account.email} (${account.linkedProviders.join(", ")})`);
```
lostintangent: What do you think about prefixing this line with "Account: " if there is a proxy? That way we display two sets of key-value pairs that are clearly distinguished.

kiliwalk:
@lostintangent Thanks for your patience and advices

cli/script/command-parser.ts
```diff
@@ -273,7 +273,10 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
             .demand(/*count*/ 1, /*max*/ 2)  // Require one non-optional and one optional argument.
             .example("login", "Logs in to the CodePush server")
             .example("login --accessKey mykey", "Logs in on behalf of the user who owns and created the access key \"mykey\"")
-            .option("accessKey", { alias: "key", default: null, demand: false, description: " Access key to authenticate against the CodePush server with, instead of providing your username and password credentials", type: "string" })
+            .example("login --proxy http://someproxy.com:455", "Logs in with the specified proxy url")
+            .option("accessKey", { alias: "key", default: null, demand: false, description: "Access key to authenticate against the CodePush server with, instead of providing your username and password credentials", type: "string" })
+            .option("proxy", { default: null, demand: false, description: "URL of the proxy server to use", type: "string" })
+            .option("noProxy", { default: false, demand: false, description: "Bypass the system-wide proxy settings", type: "boolean" })
```
dtivel: `proxy` and `noProxy` sound mutually exclusive.  Globally rename `noProxy` to `bypassProxy`.

cli/script/command-executor.ts
```diff
@@ -1288,6 +1295,19 @@ function throwForInvalidOutputFormat(format: string): void {
 function whoami(command: cli.ICommand): Promise<void> {
     return sdk.getAccountInfo()
         .then((account): void => {
-            log(`${account.email} (${account.linkedProviders.join(", ")})`);
+            log(`Account: ${account.email} (${account.linkedProviders.join(", ")})`);
+
```
lostintangent: Minor nit, but I'd prefer that we only show the "Account: " prefix if we're also showing the proxy. That way, for the users not using the proxy support, the existing behavior remains the same. But if we do show the proxy, then we label both peices of data.

cli/script/command-executor.ts
```diff
@@ -1288,6 +1295,19 @@ function throwForInvalidOutputFormat(format: string): void {
...
+            var connectionInfo = deserializeConnectionInfo();
+            if (connectionInfo.noProxy) {
+                log(`Proxy: Ignored`);
+            } else if (connectionInfo.proxy) {
+                log(`Proxy: ${connectionInfo.proxy}`);
```
lostintangent: Could we use chalk.green to colorize the labels? That way it's consistent with the other key/value pairs we display throughout the CLI.

silhouettes:
Thank you @kiliwalk! Really appreciate you taking the time to do this.

PR #211, opened 2016-04-20
Better error message for overly large package size uploads

geof90:
When the server receives a package that is over the size limit, it (oddly) returns a "Not Found" error.

This PR explicitly handles that by converting it into a friendlier error message.

Related issues:
https://github.com/Microsoft/react-native-code-push/issues/301

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -504,7 +504,7 @@ export function execute(command: cli.ICommand): Promise<void> {
 
                 case cli.CommandType.whoami:
                     return whoami(command);
-                    
+
                 default:
                     // We should never see this message as invalid commands should be caught by the argument parser.
                     throw new Error("Invalid command:  " + JSON.stringify(command));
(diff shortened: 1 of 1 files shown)


cli/script/command-executor.ts
```diff
@@ -1038,6 +1038,13 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
                 .then((): void => {
                     log("Successfully released an update containing the \"" + command.package + "\" " + (isSingleFilePackage ? "file" : "directory") + " to the \"" + command.deploymentName + "\" deployment of the \"" + command.appName + "\" app.");
                 })
+                .catch((err: Error): void => {
+                    if (err.message === "Not Found") {
+                        err.message = "Release failed, the size of the upload may be too large.";
```
lostintangent: Should we be more specific here and say that app's must be below 200mb in size? Also, the "Release failed," probably isn't necessary since it would already be clear that an error happened as part of the release.

geof90: 200mb - we might tweak this from the server side in the future

Release failed - Given the cryptic nature of this server response, we are not 100% positive that the error is a result of the size limit, hence i added a generic "release failed" and weak language like "may be too large". Do u have a suggestion on how better we can word this?

geof90:
Closing pending further investigation on whether we can control the error on the server side, may reopen if otherwise.

PR #222, opened 2016-05-05
Lowercasing "ios" directory in release-react

lostintangent:
This PR fixes [react-native-code-push#329](https://github.com/Microsoft/react-native-code-push/issues/329) by simply lowercasing the "ios" directory we use when trying to auto-detect the app's binary version. The React Native templates have always generated a lowercased "ios" directory, and therefore, we should have already been using this. On Windows and most OS X machines, it wouldn't matter either way, but as noted by the bug, on Linux machines, case  matters, and therefore, this PR enables us to work across all platforms by default. 

If someone were to change the "ios" directory to be named "iOS" or "IoS" (or whatever), and run `release-react` on a machine with a case-sensitive file-system (e.g. Linux), this PR wouldn't address that scenario. However, I don't believe this is a common scenario at all, so I decided not to worry about it prematurely.

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -823,14 +823,14 @@ function getReactNativeProjectAppVersion(platform: string, projectName: string):
     var missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
     if (platform === "ios") {
         try {
-            var infoPlistContainingFolder: string = path.join("iOS", projectName);
+            var infoPlistContainingFolder: string = path.join("ios", projectName);
             var infoPlistContents: string = fs.readFileSync(path.join(infoPlistContainingFolder, "Info.plist")).toString();
         } catch (err) {
             try {
-                infoPlistContainingFolder = "iOS";
+                infoPlistContainingFolder = "ios";
                 infoPlistContents = fs.readFileSync(path.join(infoPlistContainingFolder, "Info.plist")).toString();
             } catch (err) {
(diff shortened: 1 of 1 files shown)


cli/script/command-executor.ts
```diff
@@ -823,14 +823,14 @@ function getReactNativeProjectAppVersion(platform: string, projectName: string):
     var missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
     if (platform === "ios") {
         try {
-            var infoPlistContainingFolder: string = path.join("iOS", projectName);
+            var infoPlistContainingFolder: string = path.join("ios", projectName);
             var infoPlistContents: string = fs.readFileSync(path.join(infoPlistContainingFolder, "Info.plist")).toString();
```
shishirx34: What about making Info.plist case insensitve?

shishirx34:
LGTM!

PS: I had the fix as well and waiting on Ruben to verify with travis. Since you already sent this one out, @rub8n we can try this fix out instead.

lostintangent:
Ah sorry, I thought you were working on the WKWebView issue. I also verified that this works on Travis, which isn't unique, other than using Linux by default.

cli/script/command-executor.ts
lostintangent: This casing is always used. If anything, we have a feature request to enable specifying a different name and/or location, but if the app is using Info.plist, it's always cased this way.

PR #224, opened 2016-05-06
Add yargs check for missing deployment name.

### LINKED ISSUE #220: Missing deployment name with specified yargs options for release command returns error
Repro steps:

```
code-push release foo ./www 1.0.0 -d
Upload progress:[==================================================] 100% 0.0s
[Error]  Cannot POST /apps/foo/deployments/release
```

It looks like yargs doesn't show help if you do not specify an optional argument when you specify optional parameter which is not a boolean. I wonder if we can specify non-null as an argument for yargs option.

shishirx34:
Fixes #220 - empty option for deploymentName yargs parameter calls into the wrong route.


scottbommarito:
LGTM

lostintangent:
LGTM!

PR #228, opened 2016-05-18
Revert 8122188

geof90:
This PR reverts https://github.com/Microsoft/code-push/commit/812218858b1afc170d1addc2c625ede267d53e75 because it breaks the Cordova plugin which blindly assumes that a returned object is an update.


shishirx34:
Looks good. :shipit:

PR #239, opened 2016-05-27
PLUGIN TESTING FRAMEWORK

scottbommarito:
This adds an additional top level folder, "plugin-testing-framework", which is an npm package whose primary purpose is as a library for shared functionality between the Cordova and React-Native plugin tests. It's also easily extensible to any other plugins we may create.

To create a suite of tests, one must call `PluginTestingFramework.initializeTests` with a custom `ProjectManager` (which handles the setup behavior), a suite of tests (created with a `PluginTestingFramework.TestBuilder` object), and an array of custom `IPlatform` objects (which represent the platforms that are supported, for example Android or iOS).

A lot of this code is similar to what's currently in [the Cordova plugin](https://github.com/Microsoft/cordova-plugin-code-push) but there are a lot of changes to the structure overall. If you're confused, it may help to check out [how the Cordova plugin uses this framework](https://github.com/Microsoft/cordova-plugin-code-push/blob/1c2152048d2a09de96b11c3d281a941102efc725/test/test.ts).

TO DO: publish to npm

note: if i recall correctly, i had to make some random changes to the code to get it to compile correctly. if you're curious what a change for for, it's probably that.


cli/test/cli.ts
```diff
@@ -703,7 +703,7 @@ describe("CLI", () => {
                 assert.equal(log.args[0].length, 1);
 
                 var actual: string = log.args[0][0];
-                var expected: codePush.Package[] = [
+                var expected: cmdexec.PackageWithMetrics[] = [
```
scottbommarito: this and the changes above fix a typescript semantics bug we had that was preventing me from building

plugin-testing-framework/definitions/replace.d.ts
```diff
@@ -0,0 +1,13 @@
+interface ReplaceOptions {
+	regex: string;
+	replacement: string;
+	paths: Array<string>;
+	recursive: boolean;
+	silent: boolean;
```
geof90: I guess we can sort these

scottbommarito: i suppose but afaik this was retrieved from definitelytyped or something, so correct me if i'm wrong but there's no point in changing it because we'd have to do the same updates when a new release of replace comes out or try to merge our changes into their typings file

geof90: If its retrieved from DefinitelyTyped (I don't think it is), we should have included that file header that is on every d.ts file there

```
// Type definitions for xxx
// Project: xxx
// Definitions by: xxx
// Definitions: xxx
```

scottbommarito: ok then, but then is it ok to have like 3-4 copies of this file scattered around our repositories? should we put it on definitelytyped or something similar ourselves

geof90: I agree it's not ideal, and yes we could, you could always submit it there if you want, otherwise I think it's fine.

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+                console.log(platformName + " emulator is not ready yet!");
+                checkDeferred.reject(error);
+            });
+            
+        return checkDeferred.promise;
+    }
+    
+    var emulatorReadyAttempts = 0;
+    // Loops checks to see if the emulator is ready and eventually fails after surpassing emulatorMaxReadyAttempts.
+    function checkEmulatorReadyLooper(): Q.Promise<string> {
```
geof90: Why does this need to return `string`? It seems you are resolving with `undefined` anyway.

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+
+// bootEmulatorInternal constants
+const emulatorMaxReadyAttempts = 5;
+const emulatorReadyCheckDelayMs = 30 * 1000;
+
+/**
+ * Helper function for EmulatorManager implementations to use to boot an emulator with a given platformName and check, start, and kill methods.
+ */
+function bootEmulatorInternal(platformName: string, restartEmulators: boolean, targetEmulator: string,
+    checkEmulator: () => Q.Promise<string>, startEmulator: (targetEmulator: string) => Q.Promise<string>, killEmulator: () => Q.Promise<string>): Q.Promise<string> {
```
geof90: Why `string`? All of the internal functions in this method too.

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+     */
+    restartApplication(appId: string): Q.Promise<string> {
+        return this.endRunningApplication(appId)
+            .then<void>(() => {
+                // Wait for a second before restarting.
+                return Q.delay(1000);
+            })
+            .then<string>(() => {
+            return this.launchInstalledApplication(appId);
+        });
```
geof90: This and the line above should be indented four spaces to the right?

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+     * Navigates away from the current app, waits for a delay (defaults to 1 second), then navigates to the specified app.
+     */
+    resumeApplication(appId: string, delayBeforeResumingMs: number = 1000): Q.Promise<string> {
+        // Open a default Android app (for example, settings).
+        return this.launchInstalledApplication("com.android.settings")
+            .then<void>(() => {
+                console.log("Waiting for " + delayBeforeResumingMs + "ms before resuming the test application.");
+                return Q.delay(delayBeforeResumingMs);
+            })
+            .then<string>(() => {
```
geof90: `<void>`?

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+            
+            if (!targetIOSEmulator) {
+                // If no iOS simulator is specified, get the most recent iOS simulator to run tests on.
+                tu.TestUtil.getProcessOutput("xcrun simctl list", { noLogCommand: true, noLogStdOut: true, noLogStdErr: true })
+                    .then<string>(
+                        (listOfDevices: string) => {
+                            var phoneDevice = /iPhone (\S* )*(\(([0-9A-Z-]*)\))/g;
+                            var match = listOfDevices.match(phoneDevice);
+                            deferred.resolve(match[match.length - 1]);
+                            return undefined;
```
geof90: Why is this line even needed: `return undefined`? There seem to be a ton of places in your code that require u to explicitly call `resolve(undefined)` or `return undefined` which is weird and I wonder if they can be avoided.

plugin-testing-framework/script/projectManager.ts
```diff
@@ -0,0 +1,70 @@
...
+     * Overwrite this in your implementation!
+     */
+    public getPluginName(): string { throw ProjectManager.NOT_IMPLEMENTED_ERROR_MSG; }
+
+	/**
+	 * Creates a new test application at the specified path, and configures it
+	 * with the given server URL, android and ios deployment keys.
+     * 
+     * Overwrite this in your implementation!
+	 */
```
geof90: formatting is off?

plugin-testing-framework/script/platform.ts
scottbommarito: i used `Q.Promise<string>` everywhere as opposed to switching between `string` and `void` all the time, it made it simpler to deal with annoying typing issues.

plugin-testing-framework/script/projectManager.ts
scottbommarito: weird...too much copy-pasta to move this code around

plugin-testing-framework/script/platform.ts
scottbommarito: i used `Q.Promise<string>` everywhere when i built this code, mostly because i was having trouble with the typing and being consistent with one type seemed to resolve it (i was new to promises at the time). i used `string` specifically because the only time these promise chains return anything defined is in the TestUtil.getProcessOutput function. i could probably go through and fix it all if you think that would be significantly better.

plugin-testing-framework/script/platform.ts
geof90: Doesn't that defeat the purpose of typing? We might as well use `any` everywhere, going by the same argument (simplicity and convenience). If a return value is not expected, it should be `void`. Especially now that its a framework, the typings should give a clear indication to the user what the method expects as inputs and outputs.

scottbommarito: it was mostly because i was new to promises and was having trouble getting it work correctly lol...can fix if necessary

plugin-testing-framework/script/test.ts
```diff
@@ -0,0 +1,470 @@
...
+const TEST_RUN_DIRECTORY_OPTION_NAME: string = "--test-directory";
+const DEFAULT_TEST_RUN_DIRECTORY = path.join(os.tmpdir(), testUtil.getPluginName(), "test-run");
+
+const TEST_UPDATES_DIRECTORY_OPTION_NAME: string = "--updates-directory";
+const DEFAULT_UPDATES_DIRECTORY = path.join(os.tmpdir(), testUtil.getPluginName(), "updates");
+
+const CORE_TESTS_ONLY_OPTION_NAME: string = "--core";
+
+const PULL_FROM_NPM_OPTION_NAME: string = "--npm";
+const NPM_PLUGIN_PATH: string = "cordova-plugin-code-push";
```
geof90: wouldn't this change in the React Native tests?

scottbommarito: my bad! i guess i've never used --npm with react-native tests before so i never noticed

plugin-testing-framework/script/test.ts
```diff
@@ -0,0 +1,470 @@
...
+
+export class TestBuilder {
+    public only: boolean;
+    public skip: boolean;
+    
+    constructor(options?: { only?: boolean, skip?: boolean }) {
+        if (!options) return;
+        
+        if (!!options.only) this.only = options.only;
+        if (!!options.skip) this.skip = options.skip;
```
geof90: nit: `this.only = !!options.only`

plugin-testing-framework/script/test.ts
```diff
@@ -0,0 +1,470 @@
...
+    /** Whether or not this.testBuilders directly contains any TestBuilder objects with an only flag */
+    private hasOnly: boolean;
+    
+    /**
+     * describeName - used as the description in the call to describe
+     * scenarioPath - if specified, will be set up before the tests run
+     * testBuilders - the testBuilders to create within this describe call
+     * only - if true, use describe.only
+     */
+    constructor(describeName: string, testBuilders: TestBuilder[], scenarioPath?: string, options?: { only?: boolean, skip?: boolean }) {
```
geof90: Can we have `TestBuilder.describe(...)` and `TestBuilder.it(...)` wrapper functions that wrap around the construction of these new classes so that it mimics the mocha API as much as possible? That way contributors would still know how to work with our tests without having to learn too many new things.

scottbommarito: talked with u about this and i'm totally on board for this

plugin-testing-framework/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "code-push-plugin-testing-framework",
+  "version": "1.9.0-beta",
```
geof90: 0.0.1?

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+    // Loops checks to see if the emulator is ready and eventually fails after surpassing emulatorMaxReadyAttempts.
+    function checkEmulatorReadyLooper(): Q.Promise<string> {
+        var looperDeferred = Q.defer<string>();
+        emulatorReadyAttempts++;
+        if (emulatorReadyAttempts > emulatorMaxReadyAttempts) {
+            console.log(platformName + " emulator is not ready after " + emulatorMaxReadyAttempts + " attempts, abort.");
+            deferred.reject(platformName + " emulator failed to boot.");
+            looperDeferred.resolve(undefined);
+        }
+        setTimeout(() => {
```
geof90: whitespace after a block closing

plugin-testing-framework/script/platform.ts
```diff
@@ -0,0 +1,512 @@
...
+        return looperDeferred.promise;
+    }
+    
+    // Starts and loops the emulator.
+    function startEmulatorAndLoop(): Q.Promise<string> {
+        console.log("Booting " + platformName + " emulator named " + targetEmulator + ".");
+        startEmulator(targetEmulator).catch((error) => { console.log(error); deferred.reject(error); });
+        return checkEmulatorReadyLooper();
+    }
+    var promise: Q.Promise<string>;
```
geof90: can we add a whitespace after the end of a `{...}` block?

plugin-testing-framework/script/test.ts
```diff
@@ -0,0 +1,470 @@
+"use strict";
```
geof90: I feel like this file could be broken down better. The server related code can go in one file, the test runner in one file, and the test API (describe, it) can go into another file.

plugin-testing-framework/script/test.ts
```diff
@@ -0,0 +1,470 @@
...
+        
+        if (shouldSetup) describe("Setting Up For Tests", () => setupTests());
+        else {
+            targetPlatforms.forEach(platform => {
+                var prefix: string = (onlyRunCoreTests ? "Core Tests " : "Tests ") + thisPluginPath + " on ";
+                describe(prefix + platform.getName(), () => createAndRunTests(platform));
+            });
+        }
+    });
+}
```
geof90: I feel like this file could be broken down into smaller parts. The server related code can go in one file, the test runner in one file, and the test API (describe, it) can go into another file.

plugin-testing-framework/script/test.ts
```diff
@@ -0,0 +1,470 @@
...
+
+    if (!!targetPlatform) updateResponse.downloadURL = targetPlatform.getServerUrl() + "/download";
+    
+    // We need unique hashes to avoid conflicts.
+    if (randomHash) {
+        updateResponse.packageHash = "randomHash-" + Math.floor(Math.random() * 10000);
+    }
+    
+    return updateResponse;
+}
```
geof90: I feel like maybe the `createXXResponse` methods can go to the server util file.

plugin-testing-framework/typings/code-push-plugin-testing-framework.d.ts
```diff
@@ -0,0 +1,524 @@
+declare module 'code-push-plugin-testing-framework/script/platform' {
```
geof90: Does this file need to be checked in? It's generated right?

scottbommarito: correct

plugin-testing-framework/script/test.ts
scottbommarito: i agree

plugin-testing-framework/script/test.ts
scottbommarito: i tried to put all the stuff someone using the plugin would want in here (excluding ProjectManager and Platforms, which need to be extended from their classes), but i guess i agree with u that this stuff can be moved around, especially given that now that i think about it, "PluginTestingFramework" is both a redundant and excessively long name for an import

scottbommarito: the file has been broken down

plugin-testing-framework/script/testBuilder.ts
```diff
@@ -0,0 +1,210 @@
...
+    only(description: string, tests: TestBuilderBase[], scenarioPath?: string): TestBuilderBase;
+    skip(description: string, tests: TestBuilderBase[], scenarioPath?: string): TestBuilderBase;
+}
+
+/**
+ * Returns a hybrid type that mimics mocha's describe object.
+ */
+function getDescribe(): ITestBuilderContextDefintion {
+    var describer = <ITestBuilderContextDefintion>function(description: string, tests: TestBuilderBase[], scenarioPath?: string): TestBuilderBase {
+        return new TestBuilderDescribe(description, tests, scenarioPath);
```
geof90: Instead of returning a new instance of the "TestBuilderDescribe" or "TestBuilderIt" to the user, do you think you can push the new instance to a global array so that the user doesn't have to deal with keeping it in an array himself?

scottbommarito: i will do this

silhouettes:
We don't have the resources to finish this off, and we may be making very large architectural changes, so I'm closing this PR for now. We can re-open it later if needed!

PR #241, opened 2016-05-31
Add disabled flag in patch command to README

patoroco:


lostintangent:
@patoroco Thanks for catching this!

PR #245, opened 2016-06-02
Access key changes

geof90:
This PR:

- Adds the ability to specify an expiry to access keys. If unspecified, keys will be created with a default expiry (60 days) on the server side.
   ```
$ code-push access-key add "Temporary key" --maxAge 5m

Successfully created a new access key "Temporary key": NnPx5-nmPnooZeMJygxmcqUd6wptVJx1XOClg
(Expires: Wed Jun 01 2016 16:59:19 GMT-0700 (PDT))
   ```
- Adds the ability to edit the name and expiry of your access keys.

   ```
$ code-push access-key edit "Old name" --name "New name" --maxAge "7h 15m"
Successfully renamed the access key "Old name" to "New name" and changed its expiry to Thu Jun 02 2016 00:21:43 GMT-0700 (PDT).
   ```

- Adds the ability to see the expiry of all your keys. In an upcoming server change, the server will no longer show the actual key strings anymore when the CLI tries to list access keys, the only time the actual key string will be displayed is during creation.

   ```
$ code-push access-key ls
   ```
image

@lostintangent

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -120,7 +120,7 @@ When logging in via this method, the access key will not be automatically invali
 Finally, if you need to change a key's name or expiry date, you can use the following command:
 
 ```
-code-push access-key edit <accessKeyName> --name "new name" --ttl 10d
+code-push access-key patch <accessKeyName> --name "new name" --ttl 10d
 ```
 
 ## App Management
--- a/cli/README.md
+++ b/cli/README.md
@@ -21,7 +21,7 @@ CodePush is a cloud service that enables Cordova and React Native developers to 
 * [Viewing Release History](#viewing-release-history)
 * [Clearing Release History](#clearing-release-history)
 
-[[中文版]](./README-cn.md)
+[[Chinese version 中文版]](./README-cn.md)
 
 ## Installation
 
@@ -94,18 +94,20 @@ code-push logout
 ```
 
 If you forget to logout from a machine you'd prefer not to leave a running session on (e.g. your friend's laptop), you can use the following commands to list and remove any "live" access keys.
-The list of access keys will display the name of the machine the key was created on, as well as the time the login occurred. This should make it easy to spot keys you don't want to keep around.
+The list of access keys will display the name of the machine the key was created on, the time the login occurred, and the time it expires. This should make it easy to spot keys you don't want to keep around.
 
 ```
 code-push access-key ls
-code-push access-key rm <accessKey>
+code-push access-key rm <accessKeyName>
 ```
 
-If you need additional keys that can be used to authenticate against the CodePush service without needing to give access to your GitHub and/or Microsoft credentials, you can run the following command to create a persistent one (along with a description of what it is for):
+If you need additional keys that can be used to authenticate against the CodePush service without needing to give access to your GitHub and/or Microsoft credentials, you can run the following command to create a persistent one (along with a name describing what it is for):
 
 ```
 code-push access-key add "VSTS Integration"
 ```
+
+By default, access keys expire in 60 days. You can specify a different expiry duration by using the `--maxAge` option and passing in a [human readable duration string](https://github.com/jkroso/parse-duration#parsestr) (e.g. "2d" => 2 days, "1h 15 min" => 1 hour and 15 minutes). For security, the key will only be shown once on creation, so remember to save it somewhere if needed!
 
 After creating the new key, you can specify its value using the `--accessKey` flag of the `login` command, which allows you to perform "headless" authentication, as opposed to launching a browser.
 
@@ -114,6 +116,12 @@ code-push login --accessKey <accessKey>
 ```
 
 When logging in via this method, the access key will not be automatically invalidated on logout, and can be used in future sessions until it is explicitly removed from the CodePush server. However, it is still recommended to log out once your session is complete, in order to remove your credentials from disk.
+
+Finally, if you need to change a key's name or expiry date, you can use the following command:
+
+```
+code-push access-key edit <accessKeyName> --name "new name" --maxAge 10d
+```
 
 ## App Management
 
@@ -248,7 +256,7 @@ The install metrics have the following meaning:
 
 * **Rollbacks** - The number of times that this release has been automatically rolled back on the client. Ideally this number should be zero, and in that case, this metric isn't even shown. However, if you released an update that includes a crash as part of the installation process, the CodePush plugin will roll the end-user back to the previous release, and report that issue back to the server. This allows your end-users to remain unblocked in the event of broken releases, and by being able to see this telemetry in the CLI, you can identify erroneous releases and respond to them by [rolling it back](#rolling-back-undesired-updates) on the server.
 
-* **Rollout** - Indicates the percentage of users that are elligble to receive this update. This property will only be displayed for releases that represent an "active" rollout, and therefore, have a rollout percentage that is less than 100%. Additionally, since a deployment can only have one active rollout at any given time, this label would only be present on the latest release within a deployment. 
+* **Rollout** - Indicates the percentage of users that are elligble to receive this update. This property will only be displayed for releases that represent an "active" rollout, and therefore, have a rollout percentage that is less than 100%. Additionally, since a deployment can only have one active rollout at any given time, this label would only be present on the latest release within a deployment.
 
 * **Disabled** - Indicates whether the release has been marked as disabled or not, and therefore, is downloadable by end users. This property will only be displayed for releases that are actually disabled.
 
@@ -657,7 +665,7 @@ This is the same parameter as the one described in the [above section](#rollout-
 This is the same parameter as the one described in the [above section](#target-binary-version-parameter), and simply allows you to override the target binary version that will be used for the promoted release. If unspecified, the new release will inherit the target binary version property from the release being promoted.
 
 ```shell
-# Promote the release to production and make it 
+# Promote the release to production and make it
 # available to all versions using that deployment
 code-push promote MyApp Staging Production -t "*"
 ```
--- a/cli/README.md
+++ b/cli/README.md
@@ -256,7 +256,7 @@ The install metrics have the following meaning:
 
 * **Rollbacks** - The number of times that this release has been automatically rolled back on the client. Ideally this number should be zero, and in that case, this metric isn't even shown. However, if you released an update that includes a crash as part of the installation process, the CodePush plugin will roll the end-user back to the previous release, and report that issue back to the server. This allows your end-users to remain unblocked in the event of broken releases, and by being able to see this telemetry in the CLI, you can identify erroneous releases and respond to them by [rolling it back](#rolling-back-undesired-updates) on the server.
 
-* **Rollout** - Indicates the percentage of users that are elligble to receive this update. This property will only be displayed for releases that represent an "active" rollout, and therefore, have a rollout percentage that is less than 100%. Additionally, since a deployment can only have one active rollout at any given time, this label would only be present on the latest release within a deployment.
+* **Rollout** - Indicates the percentage of users that are eligible to receive this update. This property will only be displayed for releases that represent an "active" rollout, and therefore, have a rollout percentage that is less than 100%. Additionally, since a deployment can only have one active rollout at any given time, this label would only be present on the latest release within a deployment.
 
 * **Disabled** - Indicates whether the release has been marked as disabled or not, and therefore, is downloadable by end users. This property will only be displayed for releases that are actually disabled.
 
--- a/cli/README.md
+++ b/cli/README.md
@@ -107,7 +107,7 @@ If you need additional keys that can be used to authenticate against the CodePus
 code-push access-key add "VSTS Integration"
 ```
 
-By default, access keys expire in 60 days. You can specify a different expiry duration by using the `--maxAge` option and passing in a [human readable duration string](https://github.com/jkroso/parse-duration#parsestr) (e.g. "2d" => 2 days, "1h 15 min" => 1 hour and 15 minutes). For security, the key will only be shown once on creation, so remember to save it somewhere if needed!
+By default, access keys expire in 60 days. You can specify a different expiry duration by using the `--ttl` option and passing in a [human readable duration string](https://github.com/jkroso/parse-duration#parsestr) (e.g. "2d" => 2 days, "1h 15 min" => 1 hour and 15 minutes). For security, the key will only be shown once on creation, so remember to save it somewhere if needed!
 
 After creating the new key, you can specify its value using the `--accessKey` flag of the `login` command, which allows you to perform "headless" authentication, as opposed to launching a browser.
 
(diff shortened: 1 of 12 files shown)


cli/script/command-executor.ts
```diff
@@ -101,9 +101,36 @@ export var confirm = (): Promise<boolean> => {
...
+    return sdk.addAccessKey(command.friendlyName, command.maxAge)
         .then((accessKey: AccessKey) => {
-            log("Successfully created a new access key" + (command.description ? (" \"" + command.description + "\"") : "") + ": " + accessKey.name);
+            log(`Successfully created a new access key "${command.friendlyName}": ${accessKey.name}\n(Expires: ${new Date(accessKey.expires).toString()})`);
+        });
+}
+
+function accessKeyEdit(command: cli.IAccessKeyEditCommand): Promise<void> {
+    var willEditFriendlyName = !!command.newFriendlyName && command.oldFriendlyName !== command.newFriendlyName;
+    var willEditMaxAge = command.maxAge !== null && command.maxAge !== undefined;
```
scottbommarito: can u do `!!command.maxAge` here?

geof90: It might be 0

cli/script/command-executor.ts
```diff
@@ -623,7 +653,7 @@ function logout(command: cli.ICommand): Promise<void> {
 function formatDate(unixOffset: number): string {
     var date: moment.Moment = moment(unixOffset);
     var now: moment.Moment = moment();
-    if (now.diff(date, "days") < 30) {
+    if (Math.abs(now.diff(date, "days")) < 30) {
```
scottbommarito: are you sure this is correct? [`fromNow`](http://momentjs.com/docs/#/displaying/fromnow/) by the looks of the docs, is meant for dates in the past. if you wanted to support dates in the future, the correct method to call would be [`toNow()`](http://momentjs.com/docs/#/displaying/tonow/).

```
var dateDiff = now.diff(date, "days");
if (dateDiff >= 0 && dateDiff < 30) {
    return date.fromNow();
} else if (dateDiff > -30) {
    return date.toNow();
}
```

cli/script/command-executor.ts
scottbommarito: do u think `!!command.maxAge || command.maxAge == 0` might convey that clearer?

cli/script/command-parser.ts
```diff
@@ -42,12 +43,25 @@ function updateCheck(): void {
...
+        .example("access-key " + commandName + " \"One time key\" --maxAge 5m", "Creates a new access key with the name \"One time key\" which expires in 5 minutes")
+        .option("maxAge", { default: null, demand: false, description: "A duration string specifying the time for which the access key remains valid for use", type: "string" });
 
     addCommonConfiguration(yargs);
 }
 
+function accessKeyEdit(commandName: string, yargs: yargs.Argv): void {
+    isValidCommand = true;
+    yargs.usage(USAGE_PREFIX + " access-key " + commandName + " <description>")
+        .demand(/*count*/ 3, /*max*/ 3)  // Require exactly two non-option arguments.
```
scottbommarito: `Require exactly three non-option arguments.` ?

scottbommarito:
`// LGTM`

cli/script/command-executor.ts
geof90: It works as far as I can tell. `fromNow` and `toNow` are only different based on the sign inversion.

cli/script/command-executor.ts
geof90: Not really.. the field is editable if its not null or undefined, isn't that the simplest way to think about it?

cli/script/command-executor.ts
geof90: Try it out: https://tonicdev.com/npm/moment
```
var moment = require("moment");

console.log(moment(new Date().getTime() + 120000).fromNow()); // in 2 minutes
console.log(moment(new Date().getTime() - 120000).fromNow()); // 2 minutes ago
```

scottbommarito: oh i see...their docs are quite confusing. what i wrote is actually invalid. yours is correct.

cli/script/command-executor.ts
scottbommarito: i suppose. maybe this is some common idiom that i'm just not aware of, but i think it might be good to add a comment here then instead that explains that max age could be 0.

cli/script/command-executor.ts
```diff
@@ -101,9 +101,36 @@ export var confirm = (): Promise<boolean> => {
 }
 
 function accessKeyAdd(command: cli.IAccessKeyAddCommand): Promise<void> {
-    return sdk.addAccessKey(command.description)
+    return sdk.addAccessKey(command.friendlyName, command.maxAge)
         .then((accessKey: AccessKey) => {
-            log("Successfully created a new access key" + (command.description ? (" \"" + command.description + "\"") : "") + ": " + accessKey.name);
+            log(`Successfully created a new access key "${command.friendlyName}": ${accessKey.name}\n(Expires: ${new Date(accessKey.expires).toString()})`);
+        });
```
lostintangent: Should we add some verbiage here to ensure that devs know they need to save off the key somewhere since it won't be displayed anywhere in the CLI?

cli/script/command-executor.ts
```diff
@@ -907,15 +936,34 @@ function printAccessKeys(format: string, keys: AccessKey[]): void {
     if (format === "json") {
         printJson(keys);
     } else if (format === "table") {
-        printTable(["Key", "Time Created", "Created From", "Description"], (dataSource: any[]): void => {
-            keys.forEach((key: AccessKey): void => {
-                dataSource.push([
-                    key.name,
+        printTable(["Name", "Time Created", "Created From", "Expires"], (dataSource: any[]): void => {
```
lostintangent: I realize this isn't directly related to your PR, but should we rename "Time Created" to just "Created"? With the formatting that we use for the time strings, I think it reads more naturally.

cli/script/command-parser.ts
```diff
@@ -41,10 +42,23 @@ function updateCheck(): void {
...
     isValidCommand = true;
-    yargs.usage(USAGE_PREFIX + " access-key " + commandName + " <description>")
-        .demand(/*count*/ 3, /*max*/ 3)  // Require exactly two non-option arguments.
-        .example("access-key " + commandName + " \"VSO Integration\"", "Creates a new access key with the description \"VSO Integration\"");
+    yargs.usage(USAGE_PREFIX + " access-key " + commandName + " <accessKeyName>")
+        .demand(/*count*/ 3, /*max*/ 3)  // Require exactly three non-option arguments.
+        .example("access-key " + commandName + " \"VSTS Integration\"", "Creates a new access key with the name \"VSO Integration\" which expires by default in 60 days")
+        .example("access-key " + commandName + " \"One time key\" --maxAge 5m", "Creates a new access key with the name \"One time key\" which expires in 5 minutes")
+        .option("maxAge", { default: null, demand: false, description: "A duration string specifying the time for which the access key remains valid for use", type: "string" });
+
```
lostintangent: I can appreciate why you went with "maxAge" as the name of this, but I can't help but feel like it sounds odd. The use of "max" is pretty strange, since I'm assuming our server has no intention of expiring the key sooner than the specified durarion, unlike how a browser or intermediary server might do with an HTTP response. I think as long as our docs and inline help description explain that this value is relative to the current time, as opposed to being an absolute date, we could call it "expiration" or "ttl" or whatever, and it wouldn't be confusing.

cli/script/command-parser.ts
```diff
@@ -41,10 +42,23 @@ function updateCheck(): void {
...
+    addCommonConfiguration(yargs);
+}
 
+function accessKeyEdit(commandName: string, yargs: yargs.Argv): void {
```
lostintangent: I need to think about this a little but, but my initial feeling here is that I'd prefer to have a "rename" command (like we have for app and deployment already), and then a "renew" command that specifically refreshed the TTL of the key. The use of rename feels important for consistency, and I can't help but feel like a renew command seems more intuitive than editing a maxAge field, especially since we don't display the max age value in the access key list. We show the remaining time left.

cli/script/command-parser.ts
geof90: Initially, this field was called `expiresIn`. I changed it to `maxAge` with a suggestion on my other PR in the server repo. Do you think `expiresIn` is better?

If we do change it, I prefer to not just change it on the CLI but change the actual field that the server expects too, so its less confusing.

cli/script/command-parser.ts
geof90: I'm fine with making them two seperate commands. It seems a little strange though that you can make an access key expire earlier using the "Renew" command?

scottbommarito: yea i'd prefer a different name than `renew`, but i can't think of anything other than `changeExpiry` which is definitely too clunky

cli/README.md
```diff
@@ -248,7 +256,7 @@ The install metrics have the following meaning:
...
 
-* **Rollout** - Indicates the percentage of users that are elligble to receive this update. This property will only be displayed for releases that represent an "active" rollout, and therefore, have a rollout percentage that is less than 100%. Additionally, since a deployment can only have one active rollout at any given time, this label would only be present on the latest release within a deployment. 
+* **Rollout** - Indicates the percentage of users that are elligble to receive this update. This property will only be displayed for releases that represent an "active" rollout, and therefore, have a rollout percentage that is less than 100%. Additionally, since a deployment can only have one active rollout at any given time, this label would only be present on the latest release within a deployment.
```
dtivel: eligible

cli/definitions/cli.ts
```diff
@@ -35,7 +36,14 @@ export interface ICommand {
...
 
 export interface IAccessKeyAddCommand extends ICommand {
-    description: string;
+    friendlyName: string;
+    ttl?: number;
+}
+
+export interface IAccessKeyEditCommand extends ICommand {
+    oldFriendlyName: string;
+    newFriendlyName?: string;
```
dtivel: sort

cli/script/command-executor.ts
```diff
@@ -101,9 +101,38 @@ export var confirm = (): Promise<boolean> => {
...
-            log("Successfully created a new access key" + (command.description ? (" \"" + command.description + "\"") : "") + ": " + accessKey.name);
+            log(`Successfully created a new access key "${command.friendlyName}": ${accessKey.name}`);
+            log(`(Expires: ${new Date(accessKey.expires).toString()})`);
+            log(`Please save this key as it will only be shown once!`);
+        });
+}
+
+function accessKeyEdit(command: cli.IAccessKeyEditCommand): Promise<void> {
+    var willEditFriendlyName = isCommandOptionSpecified(command.newFriendlyName) && command.oldFriendlyName !== command.newFriendlyName;
+    var willEditTtl = isCommandOptionSpecified(command.ttl);
```
dtivel: Both `:boolean` above

cli/script/command-executor.ts
```diff
@@ -907,15 +938,34 @@ function printAccessKeys(format: string, keys: AccessKey[]): void {
...
         printJson(keys);
     } else if (format === "table") {
-        printTable(["Key", "Time Created", "Created From", "Description"], (dataSource: any[]): void => {
-            keys.forEach((key: AccessKey): void => {
-                dataSource.push([
-                    key.name,
+        printTable(["Name", "Created", "Origin", "Expires"], (dataSource: any[]): void => {
+            var now = new Date().getTime();
+
+            function isExpired(key: AccessKey) {
```
dtivel: `: boolean`

cli/script/command-executor.ts
```diff
@@ -907,15 +938,34 @@ function printAccessKeys(format: string, keys: AccessKey[]): void {
...
+                return now >= key.expires;
+            }
+
+            function keyToTableRow(key: AccessKey, dim: boolean) {
```
dtivel: `: string[]`

cli/script/command-parser.ts
```diff
@@ -41,10 +42,23 @@ function updateCheck(): void {
 
 function accessKeyAdd(commandName: string, yargs: yargs.Argv): void {
     isValidCommand = true;
-    yargs.usage(USAGE_PREFIX + " access-key " + commandName + " <description>")
-        .demand(/*count*/ 3, /*max*/ 3)  // Require exactly two non-option arguments.
-        .example("access-key " + commandName + " \"VSO Integration\"", "Creates a new access key with the description \"VSO Integration\"");
+    yargs.usage(USAGE_PREFIX + " access-key " + commandName + " <accessKeyName>")
+        .demand(/*count*/ 3, /*max*/ 3)  // Require exactly three non-option arguments.
+        .example("access-key " + commandName + " \"VSTS Integration\"", "Creates a new access key with the name \"VSO Integration\" which expires by default in 60 days")
```
dtivel: VSTS

cli/script/command-parser.ts
```diff
@@ -781,4 +819,12 @@ function getServerUrl(url: string): string {
     return url;
 }
 
+function isDefined(object: any) {
```
dtivel: `: boolean`

cli/script/command-parser.ts
```diff
@@ -781,4 +819,12 @@ function getServerUrl(url: string): string {
...
+    return object !== undefined && object !== null;
+}
+
+function parseDurationMilliseconds(durationString: string) {
```
dtivel: `: number`

cli/script/command-executor.ts
```diff
@@ -101,9 +101,38 @@ export var confirm = (): Promise<boolean> => {
...
+            log(`Please save this key as it will only be shown once!`);
+        });
+}
+
+function accessKeyEdit(command: cli.IAccessKeyEditCommand): Promise<void> {
+    var willEditFriendlyName = isCommandOptionSpecified(command.newFriendlyName) && command.oldFriendlyName !== command.newFriendlyName;
+    var willEditTtl = isCommandOptionSpecified(command.ttl);
+
+    if (!willEditFriendlyName && !willEditTtl) {
+        throw new Error("A new name or TTL must be provided.");
```
dtivel: and/or

cli/script/command-parser.ts
dtivel: Not sure you should specify the default here, since in your description you clearly say that the default is server-side and subject to change.

dtivel:
Looks good.

cli/script/command-parser.ts
geof90: Thinking about this more.. I think the user deserves to know that there is a default, and what it is. If it changes, we should update the docs. I think I will remove the part that mentions the fact that the default is server-side and subject to change.

sdk/script/management-sdk.ts
```diff
@@ -90,19 +90,23 @@ class AccountManager {
         });
     }
 
-    public addAccessKey(description: string): Promise<AccessKey> {
-        if (!description) {
-            throw new Error("A description must be specified when adding an access key.");
+    public addAccessKey(friendlyName: string, ttl?: number): Promise<AccessKey> {
```
lostintangent: I would prefer that we just call this parameter `name` instead of `friendlyName`, since that is the name we display in the CLI output.

geof90: I addressed this in the other PR. Sorry for the confusion!

cli/script/command-parser.ts
geof90: @lostintangent Are you still cool with `rename` and `renew`?

lostintangent:
LGTM!

PR #246, opened 2016-06-03
code-push session ls + rm

geof90:
(Merging into "access-key-changes" branch)

This PR:

- Adds the ability to list sessions. 
  - `access-key ls` will only show access keys that are created via `access-key add`, and will show the name of the key, the time it was created, and the expiry.
  - `session ls` will only show keys that are created via the normal login process, and will show the name of the machine associated with the session and the time it was created.
- Adds the ability to remove sessions by "machine name"
- Automatically removes the cached credentials on the temp directory if they are no longer valid (expired, etc).
- Fixed a test with the Management SDK - the statusCode was not being propagated in the returned error. Hilariously, the "methods reject the promise with status code info when an error occurs" test did not actually check the status code and was blissfully passing all this time :)
- Added new TypeScript definitions distinguishing "AccessKey"s from "Session"s
- Got rid of the "friendlyName" concept and simply replaced them with "name" globally

@lostintangent

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -24,7 +24,7 @@ var which = require("which");
 import wordwrap = require("wordwrap");
 
 import * as cli from "../definitions/cli";
-import { AccessKey, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, UpdateMetrics } from "code-push/script/types";
+import { AccessKey, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, Session, UpdateMetrics } from "code-push/script/types";
 
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
 var emailValidator = require("email-validator");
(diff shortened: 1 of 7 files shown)


cli/script/command-executor.ts
```diff
@@ -970,6 +977,17 @@ function printAccessKeys(format: string, keys: AccessKey[]): void {
     }
 }
 
+function printSessions(format: string, sessions: AccessKey[]): void {
+    if (format === "json") {
+        printJson(sessions);
+    } else if (format === "table") {
+        printTable(["Machine", "Started"], (dataSource: any[]): void => {
```
lostintangent: How about "Logged in" instead of "Started"? GitHub uses the phrase "Signed in" for this in their sessions list, and it sounds more natural than "Started", since the action that the user took to create the session was to actually login.

cli/script/command-executor.ts
```diff
@@ -1347,6 +1365,47 @@ function serializeConnectionInfo(accessKey: string, preserveAccessKeyOnLogout: b
     log(`\r\nSuccessfully logged-in. Your session file was written to ${chalk.cyan(configFilePath)}. You can run the ${chalk.cyan("code-push logout")} command at any time to delete this file and terminate your session.\r\n`);
 }
 
+function sessionList(command: cli.ISessionListCommand): Promise<void> {
+    throwForInvalidOutputFormat(command.format);
+
+    return sdk.getAccessKeys()
```
lostintangent: Do we want to add a `getSessions` method to the management SDK instead of defining this behavior in the CLI?

sdk/script/management-sdk.ts
```diff
@@ -130,6 +130,11 @@ class AccountManager {
             .then(() => null);
     }
 
+    public removeSessions(createdBy: string): Promise<void> {
```
lostintangent: Should we call this `removeSession`? I appreciate that on the server there could be multiple sessions for the same machine, but the CLI gives the impression that there is a 1:1 relationship between machine names and sessions, so if feels like the SDK should do the same thing.

cli/script/command-parser.ts
```diff
@@ -128,6 +128,26 @@ function removeCollaborator(commandName: string, yargs: yargs.Argv): void {
     addCommonConfiguration(yargs);
 }
 
+function sessionList(commandName: string, yargs: yargs.Argv): void {
+    isValidCommand = true;
+    yargs.usage(USAGE_PREFIX + " session " + commandName + " [options]")
+        .demand(/*count*/ 2, /*max*/ 2)  // Require exactly two non-option arguments.
+        .example("session " + commandName, "Lists your sessions in tabular format")
+        .example("session " + commandName + " --format json", "Lists your sessions in JSON format")
+        .option("format", { default: "table", demand: false, description: "Output format to display your access keys with (\"json\" or \"table\")", type: "string" });
```
lostintangent: The description should say "login sessions" not "access keys"

cli/script/command-executor.ts
```diff
@@ -970,6 +977,17 @@ function printAccessKeys(format: string, keys: AccessKey[]): void {
     }
 }
 
+function printSessions(format: string, sessions: AccessKey[]): void {
+    if (format === "json") {
+        printJson(sessions);
+    } else if (format === "table") {
+        printTable(["Machine", "Started"], (dataSource: any[]): void => {
+            sessions.forEach((session: AccessKey) =>
+                dataSource.push([session.createdBy, formatDate(session.createdTime)]));
```
lostintangent: Is this list already going to be sorted by machine name?

geof90: It is sorted by createdTime.. you think it should be the machine name instead?

lostintangent: Nope, sorting by created time makes sense. Thanks for confirming.

cli/script/command-executor.ts
geof90: I'm not sure. Part of me still feels like whoever consumes the SDK should be made aware that sessions and access keys are really the same thing differentiated only with the `isSession` fieldd. And in some ways, even CLI end users would still be aware of this (when the user logs in, the CLI says "enter your access key"). Also, the `removeAccessKey`, `editAccessKey` SDK functions are all currently designed to work for session keys as well. The fact that we have a special method `removeSessions` is really just a helper method to delete access keys which have `isSession` set to `true` and a `createdBy` set to the specified value. Unless we say we want to create separate SDK functions for all of those, but I'm not sure what the value is there.

lostintangent: What your explaining seems purely based on the historical decisions that we've made, as opposed to what makes the most sense from an end user perspective. The fact that session and access keys are the same is an implementation detail that has zero value to a developer. I don't feel strongly about this, but I don't see the argument for why a consumer of the SDK should know that sessions and access keys are the same.

geof90: The most tangible value is that if `getAccessKeys` returns everything, it becomes obvious to the developer that `removeAccessKey`, `editAccessKey` should work with whatever was returned from `getAccessKeys`. 

Unless we are saying that `removeAccessKey` and `editAccessKey` shouldn't work with session keys (and the fact that they in fact actually do today is not something developers is not an officially supported feature and is subject to change), then I will be happy to make this change

geof90: Ok as I typed the last line, I convinced myself :p. Will make the change u suggested.

geof90: Since we're now going with this thought that access keys and session keys should just appear completely different to the end users, I'll go ahead and just create separate TypeScript interfaces for each of them.

lostintangent:
LGTM!

PR #257, opened 2016-06-14
Adding debug command

lostintangent:
This PR introduces a new `debug` command to the CLI which allows you to easily view the CodePush debug logs for a running Android or iOS app. This provides an easier on-boarding experience when users need to diagnose what may be going wrong with our plugin, and doesn't require them to use (or learn) a separate tool in order to do it (e.g. Chrome console, OS X console, Xcode console, ADB logcat).

screen shot 2016-06-14 at 10 32 00 am

*NOTE: This doesn't currently support iOS devices, but we can add that later.*

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -17,11 +17,14 @@ import * as Q from "q";
 import * as recursiveFs from "recursive-fs";
 var rimraf = require("rimraf");
 import * as semver from "semver";
+var simctl = require("simctl");
 import slash = require("slash");
 var Table = require("cli-table");
 import * as yazl from "yazl";
 var which = require("which");
 import wordwrap = require("wordwrap");
+
+var debugCommand = require("./commands/debug");
 
 import * as cli from "../definitions/cli";
 import { AccessKey, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, Session, UpdateMetrics } from "code-push/script/types";
@@ -490,6 +493,9 @@ export function execute(command: cli.ICommand): Promise<void> {
 
                 case cli.CommandType.collaboratorRemove:
                     return removeCollaborator(<cli.ICollaboratorRemoveCommand>command);
+
+                case cli.CommandType.debug:
+                    return debugCommand(<cli.IDebugCommand>command);
 
                 case cli.CommandType.deploymentAdd:
                     return deploymentAdd(<cli.IDeploymentAddCommand>command);
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -24,12 +24,11 @@ import * as yazl from "yazl";
 var which = require("which");
 import wordwrap = require("wordwrap");
 
-var debugCommand = require("./commands/debug");
-
 import * as cli from "../definitions/cli";
 import { AccessKey, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, Session, UpdateMetrics } from "code-push/script/types";
 
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
+var debugCommand = require("./commands/debug");
 var emailValidator = require("email-validator");
 var packageJson = require("../package.json");
 var parseXml = Q.denodeify(require("xml2js").parseString);
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -4,6 +4,7 @@ import AccountManager = require("code-push");
 import * as base64 from "base-64";
 import * as chalk from "chalk";
 var childProcess = require("child_process");
+import debugCommand from "./commands/debug";
 import * as fs from "fs";
 var g2js = require("gradle-to-js/lib/parser");
 import * as moment from "moment";
@@ -23,12 +24,10 @@ var Table = require("cli-table");
 import * as yazl from "yazl";
 var which = require("which");
 import wordwrap = require("wordwrap");
-
 import * as cli from "../definitions/cli";
 import { AccessKey, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, Session, UpdateMetrics } from "code-push/script/types";
 
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
-var debugCommand = require("./commands/debug");
 var emailValidator = require("email-validator");
 var packageJson = require("../package.json");
 var parseXml = Q.denodeify(require("xml2js").parseString);
--- a/cli/script/command-parser.ts
+++ b/cli/script/command-parser.ts
@@ -260,6 +260,16 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
 
         addCommonConfiguration(yargs);
     })
+    .command("debug", "View the CodePush debug logs for a running app", (yargs: yargs.Argv) => {
+        isValidCommandCategory = true;
+        isValidCommand = true;
+        yargs.usage(USAGE_PREFIX + " debug <platform>")
+            .demand(/*count*/ 2, /*max*/ 2)
+            .example("debug android", "View the CodePush debug logs for an Android emulator or device")
+            .example("debug ios", "View the CodePush debug logs for the iOS simulator");
+
+        addCommonConfiguration(yargs);
+    })
     .command("deployment", "View and manage your app deployments", (yargs: yargs.Argv) => {
         isValidCommandCategory = true;
         yargs.usage(USAGE_PREFIX + " deployment <command>")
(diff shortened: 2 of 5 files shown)


cli/script/command-executor.ts
```diff
@@ -17,12 +17,15 @@ import * as Q from "q";
...
 var rimraf = require("rimraf");
 import * as semver from "semver";
+var simctl = require("simctl");
 import slash = require("slash");
 var Table = require("cli-table");
 import * as yazl from "yazl";
 var which = require("which");
 import wordwrap = require("wordwrap");
 
+var debugCommand = require("./commands/debug");
```
scottbommarito: put this in alphabetical order, right between `configFilePath` and `emailValidator`

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
+import * as childProcess from "child_process";
+import * as cli from "../../definitions/cli";
+import * as moment from "moment";
+import * as path from "path";
+import * as Q from "q";
+
+const simctl = require("simctl");
```
scottbommarito: do we need oss approval for this?

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+            throw new Error("iOS debug logs can only be viewed on OS X.");
+        }
+
+        const simulatorID = this.getIOSSimulatorID();
+        if (!simulatorID) {
+            throw new Error("No iOS simulators found. Re-run this command after starting one."); 
+        }
+
+        const logFilePath: string = path.join(process.env.HOME, "Library/Logs/CoreSimulator", simulatorID, "system.log");
+        return childProcess.spawn("tail", ["-f", logFilePath]);
```
scottbommarito: i'm pretty sure this will explode on windows

lostintangent: That's why I check that the `process.platform` is OS X first.

scottbommarito: my bad, im dumb lol

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+};
+
+module.exports = (command: cli.IDebugCommand): Q.Promise<void> => {
+    return Q.Promise<void>((resolve, reject) => {
+        const platform: string = command.platform.toLowerCase();
+        const debugPlatform: IDebugPlatform = debugPlatforms[platform];
+
+        if (!debugPlatform) {
+            const availablePlatforms = Object.getOwnPropertyNames(debugPlatforms);
+            return reject(new Error(`"${platform}" is an unsupported platform. Available options are ${availablePlatforms.join(", ")}.`));
```
scottbommarito: so, correct me if i'm wrong, but if you're on windows asking for ios it will let you but crash at the `process.platform !== "darwin"` check above instead of here, right? this check is only for if you asked for a nonexistent platform like "windows-phone"?

scottbommarito:
`// LGTM`

i would run it on a windows just to verify it works but it should be fine.

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const simctl = require("simctl");
+const which = require("which");
+
+interface IDebugPlatform {
+    getLogProcess(): any;
+}
+
+class AndroidDebugPlatform implements IDebugPlatform {
+    getLogProcess() {
```
shishirx34: `public getLogProcess(): any`

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+    private isDeviceAvailable() {
+        const output = childProcess.execSync("adb devices").toString();
+        return output.search(/^[\w-]+\s+device$/mi) > -1;
+    }
+}
+
+class iOSDebugPlatform implements IDebugPlatform {
+    private getSimulatorID(): string {
+        const output: any = simctl.list({ devices: true, silent: true });
+        const simulators = output.json.devices
```
shishirx34: `: string[]` ?

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        return simulators[0];
+    }
+
+    getLogProcess() {
+        if (process.platform !== "darwin") {
+            throw new Error("iOS debug logs can only be viewed on OS X.");
+        }
+
+        const simulatorID = this.getSimulatorID();
```
shishirx34: : string

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+const logMessagePrefix = "[CodePush] ";
+function processLogData(logData: Buffer) {
+    const content = logData.toString()
+    content.split("\n")
+        .filter((line) => line.indexOf(logMessagePrefix) > -1)
+        .map((line) => {
+            const timeStamp = moment().format("hh:mm:ss");
```
shishirx34: : string

cli/script/commands/debug.ts
lostintangent: Correct. This is basically doing syntax checking on the free-form "platform" parameter, not on the validity of the environment for the platform you specified wanting logs for.

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+        if (!simulatorID) {
+            throw new Error("No iOS simulators found. Re-run this command after starting one."); 
+        }
+
+        const logFilePath: string = path.join(process.env.HOME, "Library/Logs/CoreSimulator", simulatorID, "system.log");
+        return childProcess.spawn("tail", ["-f", logFilePath]);
+    }
+}
+
+const logMessagePrefix = "[CodePush] ";
```
shishirx34: I think everything below from here should just go into `script/command-executor.ts` so that all our command logic stays in the same place. And you can simply call into the function locally and export the above classes.

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+import * as Q from "q";
+
+const simctl = require("simctl");
+const which = require("which");
+
+interface IDebugPlatform {
+    getLogProcess(): any;
+}
+
+class AndroidDebugPlatform implements IDebugPlatform {
```
shishirx34: export

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+    //
+    // List of devices attached
+    // emulator-5554	device
+    private isDeviceAvailable() {
+        const output = childProcess.execSync("adb devices").toString();
+        return output.search(/^[\w-]+\s+device$/mi) > -1;
+    }
+}
+
+class iOSDebugPlatform implements IDebugPlatform {
```
shishirx34: export

shishirx34:
LGTM

cli/script/commands/debug.ts
geof90: I think this is to filter the log messages from the device, its fine being kept in here.

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+        const logFilePath: string = path.join(process.env.HOME, "Library/Logs/CoreSimulator", simulatorID, "system.log");
+        return childProcess.spawn("tail", ["-f", logFilePath]);
+    }
+}
+
+const logMessagePrefix = "[CodePush] ";
+function processLogData(logData: Buffer) {
+    const content = logData.toString()
+    content.split("\n")
+        .filter((line: string) => line.indexOf(logMessagePrefix) > -1)
```
geof90: Protip: The tilde `~` operator could come in handy here!

scottbommarito: oooh i should look this up too

cli/script/commands/debug.ts
```diff
@@ -0,0 +1,105 @@
...
+        })
+        .forEach((line: string) => console.log(line));
+}
+
+const debugPlatforms: any = {
+    android: new AndroidDebugPlatform(),
+    ios: new iOSDebugPlatform()
+};
+
+module.exports = (command: cli.IDebugCommand): Q.Promise<void> => {
```
shishirx34: Given that we are not moving this into the command-executor, may be change this to be an `export function debugCommand(command: cli.IDebugCommand): Q.Promise<void> {...}` to be consistent with the exports in our ts code and use ` import { debugCommand } from "./commands/debug" ` in command-executor.

geof90:
LGTM!

cli/script/command-parser.ts
```diff
@@ -599,6 +609,15 @@ function createCommand(): cli.ICommand {
                 }
                 break;
 
+            
+            case "debug":
+                cmd = { type: cli.CommandType.debug };
+
+                var debugCommand = <cli.IDebugCommand>cmd;
```
geof90: nit: This name clashes with the import/require you had above. Could we rename one of them to something else?

lostintangent: I simply followed the existing pattern used within this file, but I'm actually not sure why we're even creating a new runtime variable simply to appease TypeScript for casting purposes. This seems like a cleaner solution and resolves your comment as well:

```javascript
cmd = <cli.IDebugCommand>{
    type: cli.CommandType.debug,
    platform: arg1
};
```

cli/script/commands/debug.ts
lostintangent: Done!

Issue #207, opened 2016-04-15
Support for Info.plist files with different names


dudeinthemirror:
In my project, I have:
```
MyApp.PROD-Info.plist
MyApp.DEV-Info.plist
MyApp.STAGING-Info.plist
```
corresponding to the 3 targets: MyApp.PROD, MyApp.DEV and MyApp.STAGING
I can't find a way to pass the plist file name when running `code-push react-native`
So, I am getting:
```
[Error]  Unable to find or read "Info.plist" in the "iOS/MyApp" or "iOS" folders.
```
The command I am running is:
```
/usr/local/bin/code-push release-react MyApp-iOS iOS \
--deploymentName STAGING \
--description "some description" \
--entryFile index.ios.js \
--mandatory
```
Thanks

lostintangent:
@dudeinthemirror This is great feedback! We can definitely discuss adding this to support this scenario. In the meantime, the workaround is to specify the "--targetBinaryVersion" flag with the version from the respective plist file (e.g. "1.2.3"). The only reason that the release-react command tries to load the info.plist file is to infer the app version, so if you specify it explicitly, it won't even attempt to read the file and you won't get this error.

dudeinthemirror:
Thanks for the quick turnaround @lostintangent !
So actually that is one thing I like about _not_ specifying the --targetBinaryVersion i.e the fact that it just takes it from the app :)

So what I ended up doing is I added a step in the build script that does this:
```
step="Hack: creating Info.plist as a copy of ${TARGET}-Info.plist"
echo "-----------------------------------------------------------------------------------------------"
echo "STEP: $step"
echo "-----------------------------------------------------------------------------------------------"
# This is a hack until I get an answer to :
# https://github.com/Microsoft/code-push/issues/207
PLIST_FILE=$(find . -name ${TARGET}-Info.plist -print 2>/dev/null)
[[ ! -z "$PLIST_FILE" ]] && cp $PLIST_FILE ./ios/Info.plist
```
Hope it will help others who encounter this error !

I strongly suggest you guys consider adding plist-file-name as an option. Xcode allows you to rename your Info.plist as needed, so in most large projects that will be the norm.
Thank you for a great tool!

lostintangent:
That workaround would do as well :) I definitely agree that we should add this option, I just wanted to make sure you could get unblocked in the meantime. I'll use this issue to track this enhancement, since like you mentioned, this will definitely be a common need that I'd like to address ASAP.

[...]

PR #261, opened 2016-06-19
release-react plist customizations

lostintangent:
This addresses #207 by adding two new optional parameters to the `release-react` command: `plistFile` and `plistFilePrefix`. The latter allows simply augmenting the existing behavior in situations where you have created per-environment plist files (e.g. `DEV-Info.plist` vs. `PROD-Info.plist`), and you just need to let CodePush know which prefix you want it to use in order to infer the correct app version of your update. This allows devs a bit more flexibility in how they organize their app's config, without requiring them to drop down to explicitly specifying the `--targetBinaryVersion` parameter when calling `release-react` (which feels like a failure on the command's part).

The `plistFile` parameter is for situations where your plist file is in an entirely arbitrary location and/or isn't using the standard `Info.plist` suffix (not common but possible). This has the benefit of providing maximum flexibility to the app dev, while still preventing them from having to explicitly specify the target binary version.

Therefore, the options for iOS are as follows:

1. If your app is a standard React Native app, then the `release-react` command will find your `Info.plist` correctly as it already does today. This is the simplest and most common solution.

2. If you've created per-environment plist files on top of the standard React Native app template, then you can optionally specify your environment-specific prefix when calling the `release-react` command via the new `--plisrtFilePrefix` parameter.

3. If your app has a "non-standard" layout (e.g. its a native iOS app with embedded RN views), then you can specify the exact path to your plist file via the new `--plistFile` parameter. This is the least common yet most flexible solution.

*NOTE: If you want to specify an explicit target binary version, then you can pass the `--targetBinaryVersion` parameter to the `release-react` command, which effectively "disables" the logic which tries to find a plist file and read the version from it.*

```shell
# Release an update using a plist file that is located in a
# "well-known" directory, and simply includes a custom prefix
code-push release-react foo ios --pre "DEV-"

# Release an update using a plist file in an arbitrary
# location and with an arbitrary file name
code-push release-react foo ios -p "../config/foo.plist"
```

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -857,35 +857,52 @@ function getPackageMetricsString(obj: Package): string {
     return returnString;
 }
 
-function getReactNativeProjectAppVersion(platform: string, projectName: string): Promise<string> {
-    var missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
+function getReactNativeProjectAppVersion(platform: string, projectName: string, options: any): Promise<string> {
+    const missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
+
     if (platform === "ios") {
-        try {
-            var infoPlistContainingFolder: string = path.join("ios", projectName);
-            var infoPlistContents: string = fs.readFileSync(path.join(infoPlistContainingFolder, "Info.plist")).toString();
-        } catch (err) {
-            try {
-                infoPlistContainingFolder = "ios";
-                infoPlistContents = fs.readFileSync(path.join(infoPlistContainingFolder, "Info.plist")).toString();
-            } catch (err) {
-                throw new Error(`Unable to find or read "Info.plist" in the "ios/${projectName}" or "ios" folders.`);
-            }
-        }
-
-        try {
-            var parsedInfoPlist: any = plist.parse(infoPlistContents);
-        } catch (err) {
-            throw new Error(`Unable to parse the "${infoPlistContainingFolder}/Info.plist" file, it could be malformed.`);
-        }
-
-        if (parsedInfoPlist && parsedInfoPlist.CFBundleShortVersionString) {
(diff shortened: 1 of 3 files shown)


cli/script/command-executor.ts
```diff
@@ -857,35 +857,53 @@ function getPackageMetricsString(obj: Package): string {
     return returnString;
 }
 
-function getReactNativeProjectAppVersion(platform: string, projectName: string): Promise<string> {
-    var missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
+function getReactNativeProjectAppVersion(platform: string, projectName: string, options: any): Promise<string> {
```
geof90: I would prefer if you simply passed the `command` object or declared a new interface for `options` instead of declaring it as `any`

cli/script/command-executor.ts
```diff
@@ -857,35 +857,53 @@ function getPackageMetricsString(obj: Package): string {
...
+        if (parsedPlist && parsedPlist.CFBundleShortVersionString) {
+            if (semver.valid(parsedPlist.CFBundleShortVersionString) || missingPatchVersionRegex.test(parsedPlist.CFBundleShortVersionString)) {
+                return Q(parsedPlist.CFBundleShortVersionString);
             } else {
-                throw new Error(`The "CFBundleShortVersionString" key in "${infoPlistContainingFolder}/Info.plist" needs to have at least a major and minor version, for example "2.0" or "1.0.3".`);
+                throw new Error(`The "CFBundleShortVersionString" key in "${resolvedPlistFile}" needs to have at least a major and minor version, for example "2.0" or "1.0.3".`);
             }
         } else {
-            throw new Error(`The "CFBundleShortVersionString" key does not exist in "${infoPlistContainingFolder}/Info.plist".`);
+            throw new Error(`The "CFBundleShortVersionString" key does not exist in "${resolvedPlistFile}" file.`);
```
geof90: in the "${resolvedPlistFile}" file

geof90:
LGTM!

cli/script/command-executor.ts
lostintangent: OK, sounds good to me.

PR #263, opened 2016-06-19
Adding gradle.properties support to release-react

lostintangent:
This PR addresses https://github.com/Microsoft/react-native-code-push/issues/353. It enhances the existing Android support in the `release-react` command by allowing an app to reference a Gradle property as the value of the `versionName` setting in their `build.gradle` file, as opposed to only being able to specify a string literal. This change allows referencing the variable with or without the `project.` prefix, and allows the variable to be defined in either the app or project-wide properties file for added flexibility.

**Example build.gradle file:**

```groovy
android {
    defaultConfig {
        versionName VERSION_NAME
    }
}
```

**Corresponding gradle.properties file:**

```properties
VERSION_NAME=1.2.3
```

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -860,7 +860,8 @@ function getPackageMetricsString(obj: Package): string {
 
 function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, projectName: string): Promise<string> {
     const missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
-    
+    const validVersion = (version: string) => semver.valid(version) || missingPatchVersionRegex.test(version);
+
     const fileExists = (file: string): boolean => {
         try { return fs.statSync(file).isFile() }
         catch (e) { return false }
@@ -908,12 +909,10 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
             throw new Error(`The "CFBundleShortVersionString" key doesn't exist within the "${resolvedPlistFile}" file.`);
         }
     } else if (command.platform === "android") {
-        var buildGradlePath: string = path.join("android", "app", "build.gradle");
+        const buildGradlePath: string = path.join("android", "app", "build.gradle");
         if (fileDoesNotExistOrIsDirectory(buildGradlePath)) {
             throw new Error("Unable to find or read \"build.gradle\" in the \"android/app\" folder.");
         }
-
-        const validVersion = (version: string) => semver.valid(version) || missingPatchVersionRegex.test(version);
         
         return g2js.parseFile(buildGradlePath)
             .catch(() => {
@@ -931,12 +930,18 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
                 let appVersion: string = buildGradle.android.defaultConfig.versionName.replace(/"/g, "").trim();
 
                 if (validVersion(appVersion)) {
+                    // The versionName property is a valid semver string,
+                    // so we can safely use that and move on.
                     return appVersion;
+                } else if (Number(appVersion[0])) {
+                    // The versionName property isn't a valid semver string,
+                    // but it starts with a number, and therefore, it can't
+                    // be a valid Gradle property reference.
+                    throw new Error(`The "android.defaultConfig.versionName" property in "android/app/build.gradle" needs to specify a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).`);
                 }
 
                 // The version property isn't a valid semver string
                 // so we assume it is a reference to a property variable.
-
                 const propertyName = appVersion.replace("project.", "");
                 const propertiesFileName = "gradle.properties";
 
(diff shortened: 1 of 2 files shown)


scottbommarito:
`// LGTM`

cli/script/command-executor.ts
```diff
@@ -912,25 +913,57 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
...
+                let appVersion: string = buildGradle.android.defaultConfig.versionName.replace(/"/g, "").trim();
+
+                if (validVersion(appVersion)) {
+                    return appVersion;
                 }
+
+                // The version property isn't a valid semver string
+                // so we assume it is a reference to a property variable.
+
+                const propertyName = appVersion.replace("project.", "");
```
shishirx34: Is it a given that the app version will always be referenced as `project.APP_VERSION` in a build gradle file? Can it not be something else eg: `build.APP_VERSION`? How about just extracting the last property by splitting on `.`?

lostintangent: The `project` object is the implicit "global" in a Gradle build file, so you can reference a global property with or with the `project.` prefix, very similarly to the `window` object in the browser.

cli/script/command-executor.ts
```diff
@@ -907,30 +908,66 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
...
+                ];
+
+                const propertiesFile: string = (<any>knownLocations).find(fileExists);
+                const propertiesContent: string = fs.readFileSync(propertiesFile).toString();
+
+                try {    
+                    const parsedProperties: any = properties.parse(propertiesContent);
+                    appVersion = parsedProperties[propertyName];
+                } catch (e) {
+                    throw new Error(`Unable to parse ${propertiesFile}. Please ensure it as well-formed properties file.`);
```
shishirx34: Please ensure it __is a__ well-formed...

cli/script/command-executor.ts
lostintangent: Also, the properties that are defined via `grade.properties` are exposed to the build file as properties on the `project` object, so I don't think we need to make this solution any more generic, since there isn't another global object that property references could be made off of (e.g. `build.FOO`).

cli/script/command-executor.ts
```diff
@@ -907,30 +908,66 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
...
+                } else if (Number(appVersion[0])) {
+                    // The versionName property isn't a valid semver string,
+                    // but it starts with a number, and therefore, it can't
+                    // be a valid Gradle property reference.
+                    throw new Error(`The "android.defaultConfig.versionName" property in "android/app/build.gradle" needs to specify a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).`);
                 }
+
+                // The version property isn't a valid semver string
+                // so we assume it is a reference to a property variable.
+                const propertyName = appVersion.replace("project.", "");
```
shishirx34: My comment was overridden by the commit, re-posting:

Is it a given that the app version will always be referenced as project.APP_VERSION in a build gradle file? Can it not be something else eg: build.APP_VERSION or properties.APP_VERSION? How about just extracting the last property by splitting on .?

lostintangent: I replied to your original comment :)

cli/script/command-executor.ts
```diff
@@ -907,30 +908,66 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
...
+                    throw new Error(`The "android.defaultConfig.versionName" property value in "android/app/build.gradle" is not a valid string. If this is expected, consider using the --targetBinaryVersion option to specify the value manually.`);
+                }
+
+                let appVersion: string = buildGradle.android.defaultConfig.versionName.replace(/"/g, "").trim();
+
+                if (validVersion(appVersion)) {
+                    // The versionName property is a valid semver string,
+                    // so we can safely use that and move on.
+                    return appVersion;
+                } else if (Number(appVersion[0])) {
```
shishirx34: I don't know if its likely that appVersion starts with a letter, but this would throw if it did. 

Number(a) => Uncaught ReferenceError.

shishirx34: Wouldn't doing a regex check be safer to validate if appVersion starts with a number? if for some reason, although least likely, appVersion[0] = 0, this would be false.

`/^\d.*/.test(appVersion[0])`

shishirx34:
Looks good.

cli/script/command-executor.ts
lostintangent: Ah yeah that's a good point. I was trying to go with the simplest solution, but I forgot about the leading zero problem. Thanks for the catch!

cli/script/command-executor.ts
```diff
@@ -858,14 +859,14 @@ function getPackageMetricsString(obj: Package): string {
 }
 
 function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, projectName: string): Promise<string> {
-    const missingPatchVersionRegex: RegExp = /^\d+\.\d+$/;
+    const fileExists = (file: string): boolean => {
+        try { return fs.statSync(file).isFile() }
+        catch (e) { return false }
+    };
 
+    const validVersion = (version: string) => semver.valid(version) || /^\d+\.\d+$/.test(version);
```
geof90: `(version: string): boolean`

geof90:
LGTM!

PR #264, opened 2016-06-20
Command description tweaks

lostintangent:
This PR simply makes a few small tweaks to the descriptions for some of the access-key commands.

### DIFF
--- a/cli/script/command-parser.ts
+++ b/cli/script/command-parser.ts
@@ -44,9 +44,9 @@ function accessKeyAdd(commandName: string, yargs: yargs.Argv): void {
     isValidCommand = true;
     yargs.usage(USAGE_PREFIX + " access-key " + commandName + " <accessKeyName>")
         .demand(/*count*/ 3, /*max*/ 3)  // Require exactly three non-option arguments.
-        .example("access-key " + commandName + " \"VSTS Integration\"", "Creates a new access key with the name \"VSTS Integration\" which expires by default in 60 days")
-        .example("access-key " + commandName + " \"One time key\" --ttl 5m", "Creates a new access key with the name \"One time key\" which expires in 5 minutes")
-        .option("ttl", { default: null, demand: false, description: "A duration string specifying the time for which the access key remains valid for use", type: "string" });
+        .example("access-key " + commandName + " \"VSTS Integration\"", "Creates a new access key with the name \"VSTS Integration\", which expires in 60 days")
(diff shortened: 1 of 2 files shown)


geof90:
LGTM!

cli/script/command-parser.ts
```diff
@@ -44,9 +44,9 @@ function accessKeyAdd(commandName: string, yargs: yargs.Argv): void {
...
         .demand(/*count*/ 3, /*max*/ 3)  // Require exactly three non-option arguments.
-        .example("access-key " + commandName + " \"VSTS Integration\"", "Creates a new access key with the name \"VSTS Integration\" which expires by default in 60 days")
-        .example("access-key " + commandName + " \"One time key\" --ttl 5m", "Creates a new access key with the name \"One time key\" which expires in 5 minutes")
-        .option("ttl", { default: null, demand: false, description: "A duration string specifying the time for which the access key remains valid for use", type: "string" });
+        .example("access-key " + commandName + " \"VSTS Integration\"", "Creates a new access key with the name \"VSTS Integration\", which expires in 60 days")
+        .example("access-key " + commandName + " \"One time key\" --ttl 5m", "Creates a new access key with the name \"One time key\", which expires in 5 minutes")
+        .option("ttl", { default: "60d", demand: false, description: "Duration string which specifies the amount of time that the access key should remain valid for (e.g 5m, 60d, 1y)", type: "string" });
```
scottbommarito: does the amount of time support specifying a length in months, like `2mo` or `3mo`? if it does, might be useful to put here and below in your `(e.g...`

scottbommarito: also, this is totally minor, but given that we're using the not-immediately obvious abbreviation `ttl`, i would like to see some small indication of what it stands for somewhere in here, but i'm not sure the best way to do that

scottbommarito:
`// LGTM`, just some super minor comments

shishirx34:
LGTM

cli/script/command-parser.ts
geof90: Months have to be spelt in full (e.g. '6months'), and we already linked to how to craft this string in the docs, I don't think we have to show every single example here.

[Hmm TTL is a super commonly used term](http://lmgtfy.com/?q=ttl&l=1), I don't really think it needs spelling out in full..

lostintangent: Yeah my intention here wasn't to outline all possible values, but rather, to provide some examples which illustrate the syntax. Also, I'm comfortable with the use of TTL here, which is why we named the parameter that in the first place.

PR #265, opened 2016-06-21
Improving release-react

lostintangent:
This PR improves the `release-react` command in the following ways:

1. It adds a new "logging step" which explicitly displays the value that is inferred, and from which file, when auto-detecting the target binary version. This way, it's clearer what exactly the command is doing.

2. Updates the new `--plistFilePrefix` parameter by auto-appending a "-" if the provided value doesn't already end in a separator (currently a period or hyphen). In practice, 100% of the files I've seen use a hyphen as a separator, so this change simply conforms to the common convention.

3. Makes various improvements to the existing log messages

After this PR, I think the `release-react` command is in a pretty awesome state. However, I can imagine the need for us to add a new `--gradlePath` parameter to allow specifying where the `build.gradle` file is in apps that use a non-traditional project layout. We can address that in a future PR if we hear requests for it.

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -878,7 +878,8 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
             }
         } else {
             // Allow the plist prefix to be specified with or without a trailing
-            // seperator character, but prescribe the use of a "-" when omitted.
+            // separator character, but prescribe the use of a hyphen when omitted,
+            // since this is the most commonly used convetion for plist files.
             if (command.plistFilePrefix && /.+[^-.]$/.test(command.plistFilePrefix)) {
                 command.plistFilePrefix += "-";
             }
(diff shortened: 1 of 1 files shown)


cli/script/command-executor.ts
```diff
@@ -875,6 +877,12 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
                 throw new Error("The specified plist file doesn't exist. Please check that the provided path is correct.");
             }
         } else {
+            // Allow the plist prefix to be specified with or without a trailing
+            // seperator character, but prescribe the use of a "-" when omitted.
```
scottbommarito: `separator`

(not to be rude, but i see you misspell `separate`,`separated`, etc a lot, so you may want to make a mental note that it's `separ-` and not `seper-`)

scottbommarito:
`// LGTM`

PR #279, opened 2016-07-08
Update year in footer

yannickoo:
I have adjusted the year in the footer because it still says 2015 which is actually not correct :confused:


geof90:
Awesome! Thanks for the fix!

Issue #286, opened 2016-07-26
Release command may interpret version as a number


jkrems:
Given the following command:

```bash
code-push release FooBar scripts 8.10
```

The CLI will treat the last argument as "8.1" because the implicit number conversion kicks in. I've seen that most flags already set the type to string so I assume it only affects a few commands. But maybe it would still be a good idea to either require a `v` prefix ("v8.10") or an explicit ".0" to prevent future bugs connected to this..?

silhouettes:
Thanks for catching this! Fortunately it only seems to apply to the `release` command, and not the `release-react` and `release-cordova` commands, which might explain why we haven't seen it yet. I've just put up a fix for the root cause (the number conversion), but let me know if you'd still like to see the other changes you mentioned!

jkrems:
Thanks!

> I've just put up a fix for the root cause (the number conversion), but let me know if you'd still like to see the other changes you mentioned!

Nah, the rest of the issue was just me being paranoid about it. 👍

geof90:
While @silhouettes 's change is going in and before we release to NPM, could you try wrapping the value in quotes as a temporary workaround? So `code-push release FooBar scripts "8.10"`

jkrems:
Since according to the docs "8.10" and "8.10.0" are treated the same, we're just using "8.10.0". I *think* bash quoting doesn't make it into the node process:

```
> node -p 'process.argv' -- foo "bar"
[ '/Users/jkrems/.nvm/versions/node/v4.4.7/bin/node',
  'foo',
  'bar' ]
```

silhouettes:
This was fixed with release `1.12.3-beta`, thanks for filing this!

PR #287, opened 2016-07-27
Turn off implicit number conversion of non-hyphenated yargs arguments

silhouettes:
This is a fix for #286, where an app store version specified on the command line like `1.10` would be converted by yargs to a number, i.e. `1.1`. This only applies to non-hyphenated parameters, and not flagged options, where the type is explicitly specified.

As this is somewhat unexpected, I've turned this number conversion off for all non-hyphenated arguments. For our existing commands, I've checked that all non-hyphenated arguments are strings, so there should be no side effects.

### DIFF
--- a/cli/script/command-parser.ts
+++ b/cli/script/command-parser.ts
@@ -84,7 +84,8 @@ function accessKeyRemove(commandName: string, yargs: yargs.Argv): void {
 
 function addCommonConfiguration(yargs: yargs.Argv): void {
     yargs.wrap(/*columnLimit*/ null)
-        .strict()  // Validate hyphenated (named) arguments.
+        .string("_")                         // Interpret non-hyphenated arguments as strings (e.g. an app version of '1.10').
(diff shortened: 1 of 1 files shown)


shishirx34:
Looks Good.

geof90:
Nice!

Do you know if this comment in the README is still valid? I don't think your change would affect it, just would be nice to confirm.

> NOTE: If your semver expression starts with a special shell character or operator such as >, ^, or * , the command may not execute correctly if you do not wrap the value in quotes as the shell will not supply the right values to our CLI process. Therefore, it is best to wrap your targetBinaryVersion parameter in double quotes when calling the release command, e.g. code-push release MyApp updateContents ">1.2.3".

silhouettes:
Yup, the comment is still valid - that happens at the shell itself, rather than at yargs. I tried just then with the semver range `>1.10` and the shell responded `output is not a tty`

Issue #266, opened 2016-06-21
Allow specifying a build.gradle path to the release-react command


lostintangent:
Now that we've added the ability to customize the name/location of the `Info.plist` file when running the `release-react` command, it makes sense to introduce the equivalent capability on the Android side, so that apps with "non-standard" project layouts can continue to use the `release-react` command, without needing to explicit specify the `--targetBinaryVersion` parameter.

```shell
code-push release-react MyApp android --gradleFIle "../some/path/build.gradle"
```

*NOTE: Since `build.gradle` is a required file name, it may be nice to allow this new parameter to omit it, so that you could either specify the full path to the file, or simply the directory that contains it.*

geof90:
This has been addressed in https://github.com/Microsoft/code-push/pull/292

PR #292, opened 2016-08-03
Allow specifying a build.gradle to the release-react cmd

mtunique:
#266

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -843,7 +843,7 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
     };
 
     const isValidVersion = (version: string): boolean => !!semver.valid(version) || /^\d+\.\d+$/.test(version);
-        
+
     log(chalk.cyan(`Detecting ${command.platform} app version:\n`));
 
     if (command.platform === "ios") {
@@ -869,7 +869,7 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
                 path.join(iOSDirectory, projectName, plistFileName),
                 path.join(iOSDirectory, plistFileName)
             ];
-            
+
             resolvedPlistFile = (<any>knownLocations).find(fileExists);
 
             if (!resolvedPlistFile) {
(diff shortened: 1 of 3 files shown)


cli/script/command-executor.ts
```diff
@@ -896,11 +896,14 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
...
         }
     } else if (command.platform === "android") {
-        const buildGradlePath: string = path.join("android", "app", "build.gradle");
+        let buildGradlePath: string = path.join("android", "app", "build.gradle");
+        if (command.gradleFile) {
+            buildGradlePath = command.gradleFile;
+        }
         if (fileDoesNotExistOrIsDirectory(buildGradlePath)) {
-            throw new Error(`Unable to find the "build.gradle" file in your "android/app" directory.`);
+            throw new Error(`Unable to find gradle file "${buildGradlePath}"".`);
```
geof90: Unable to find the gradle file at "${buildGradlePath}"

You also had an extra `"`

cli/script/command-executor.ts
```diff
@@ -896,11 +896,14 @@ function getReactNativeProjectAppVersion(command: cli.IReleaseReactCommand, proj
             throw new Error(`The "CFBundleShortVersionString" key doesn't exist within the "${resolvedPlistFile}" file.`);
         }
     } else if (command.platform === "android") {
-        const buildGradlePath: string = path.join("android", "app", "build.gradle");
+        let buildGradlePath: string = path.join("android", "app", "build.gradle");
```
geof90: It would be a plus point if we could also address this requirement that @lostintangent suggested in https://github.com/Microsoft/code-push/issues/266:

> NOTE: Since build.gradle is a required file name, it may be nice to allow this new parameter to omit it, so that you could either specify the full path to the file, or simply the directory that contains it.

geof90:
Thanks for the contribution!

cli/script/command-executor.ts
mtunique: I am so sorry for this.

cli/script/command-executor.ts
mtunique: There are two parameters to handler the pList file, maybe gradle file be same.

mtunique: `plistFile` and `plistFilePrefix`.

geof90: Well, in the case of the plistFile, it is actually a prefix (e.g. STAGING-Info.plist), but for the gradle file, since the name "build.gradle" is required, there's really no value to the user in typing it vs not typing it, so I would prefer if the same option could be smart enough to support both methods of specifying the path (the actual file path or just the containing folder).

mtunique: Thanks. I got it.

mtunique:
@geof90 Now I am implement it.

geof90:
@mtunique Looks awesome! Thanks so much!

PR #296, opened 2016-08-07
Update gradle-to-js dependency and bump package.json version

geof90:
This is to include [this change](https://github.com/ninetwozero/gradle-to-js/pull/3) that allows gradle files that include function definitions to be parsed correctly.


shishirx34:
Looks good.

silhouettes:
Looks good!

PR #301, opened 2016-08-17
Add hash utilities

silhouettes:
For the upcoming code signing feature, we need a consistent hashing mechanism for directories of files. Also add unit tests, and fix the existing unit tests.


geof90:
LGTM!

Issue #307, opened 2016-08-26
Code Signing Feature


andrewwebber:
I have not spent to much time looking deep into the however my intuition is that a security enhancement could be in order.

I would like to see a workflow utilising signing (RSA or HMAC, etc.) and or encryption.
1. When an app is created via code-push cli a client side key is created for the app.
2. This key is used to at least sign the bundle 
3. An additional key is used to verify the bundle before in "getJSBundleFile"

This way the app developer can validate that the app bundle has not been modified by code push or some external entity.

This would be very attractive because without such a feature, within a thread analysis, developers need to explain to management that Microsoft can modify the app.

This would also reduce the likely-hood that developers will just try and re-implement the code push backend just to comply with security regulations.

The alternative of course would be to open source the code-push azure web app.

silhouettes:
Hi @andrewwebber, this is being actively worked on and approaching completion! ETA is 1 week. I'll leave this open, but also link it to #233.

The current process I'm implementing is like this:

1. Ensure dev has procured a `{ <privateKey>, <publicKey> }` pair. Could be a .p12/.pem + .cer, or a KeyStore, or any number of things.
1. Ensure the dev has the `privateKey` on their dev machine, and the `publicKey` baked natively into their app store release.
1. When a release is made from the CLI, the directory contents are hashed. Basically: create a sorted array of key value pairs (one for each file in the release), of the format `<relativeFilePath>: <SHA256Hash>`. Stringify this array and hash that to get a single `directoryHash`.
1. Construct a [JSON Web Token](https://jwt.io/), with the claim `contentHash: <directoryHash>`. Sign it using RS256 with the developer's private key from the CLI locally on their machine.
1. Write the JWT to a file called `.codepushrelease` at the root of the package, before it is released to the CodePush server.
1. When the mobile app receives an update, the CodePush SDK will look for the `.codepushrelease` file, read the JWT out of it, verify it against the configured public key, and then hash the contents of the update and verify it against the hash on the JWT. The update will only be installed if this verification is successful.

Let me know if you have any thoughts or criticisms on this! I'm also looking for ideas on how to achieve (1) and (2). For example, what does the configuration and release process look like? Should the same private key be used across your whole account, or per app, or per deployment? Should dev/staging builds be signed, maybe with a common debug key?

andrewwebber:
This is more than excellent news

[...]

Issue #233, opened 2016-05-25
FAQ does not answer who controls the update process.


alexanderkjeldaas:
Is the update guarded by a digital signature that I control?  What is the security model?

lostintangent:
The CLI and client SDKs rely on TLS to prevent man-in-the-middle attacks / tampering when releasing and syncing updates with the server, but we don't currently verify that updates are signed before installing them on the end user's device. That said, we're literally about to implement this exact feature to better ensure end-to-end security.

As long as your CodePush credentials aren't compromised, no one else can release updates to your apps, but the signature verification will help further mitigate issues in the event that the CodePush service itself was breached. This enhancement is one of our key items to complete leading up to the service exiting beta.

alexanderkjeldaas:
Thanks for the clarification!  And thank you for focusing on this.

lostintangent:
I'm going to close this issue for now, and we'll update the FAQ and introduce comprehensive security docs after we land the update signing capability in the next few weeks. If you want to keep this open for tracking purposes, please feel free to re-open it, otherwise, I'll ping you directly when we have an update. Thanks!

corintho:
I was looking through security information all around the documentation and could not find anything specific as was requested here. Could someone perhaps point me to the right direction, since the issue is already closed I assume I missed the docs somehow. Thanks.

silhouettes:
My sincere apologies, but we have been significantly delayed. Are you just looking for security documentation, or the update signing feature specifically? I am actively working on the latter at the moment, and the ETA is 1-2 weeks, at which point I can update you here.
__________________

@corintho, @alexanderkjeldaas 

If you don't mind, I'd be interested in gathering your feedback on the design of this feature. We would ideally like to make this compulsory for all users, but we'd also like to ensure that the process is as unobtrusive as possible.

Do any of the following appeal to you?

**Signature generation**

1. Delegate the generation of public/private key pair to the developer
2. Have a utility built into the CodePush CLI that generates it for you
3. Re-use the same credentials that you use to perform app store releases
4. Other?

(3) seemingly makes the most sense, as it requires the least work, means less keys to manage, and means that App Store releases and CodePush releases are guarded by the same credential.

**Signature storage**

1. Cache it within the CLI itself (what granularity? per app? per deployment?)
2. Manually supply it to the CLI each time you make a release
3. Sign the package yourself independently of the CLI

(1) seemingly makes the most sense to me, as it is the most convenient.

corintho:
[...]

PR #309, opened 2016-08-29
Add option to sign releases with a private key

silhouettes:
To use this feature, devs will need to:
1. Run `ssh-keygen` to generate a private/public key pair
2. Configure the CodePush SDK in their app with the public key file (PR upcoming), and perform an app store release
3. Perform CLI releases with the `-k` argument pointing to their private key file

Still to do:
1. Test on Windows
2. Update `README.md`
3. Get OSS approval for new dependencies

Potential future enhancements:
1. Support password-protected private key files
2. Refining the release hook mechanism, e.g. allowing aspects of behavior to be overriden
3. Allow hashing to be turned on independently to signing
4. Leverage KeyStore files and/or OSX keychain, and/or existing Android/iOS signing credentials
5. Cache public key or fingerprint on server for verification that releases are signed correctly
6. Key expiry
7. Certificates
8. Signing additional information beyond just the hash of the release, for example the app/deployment it was released on, the targeted binary versions, etc.

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -25,6 +25,7 @@ import * as yazl from "yazl";
 var which = require("which");
 import wordwrap = require("wordwrap");
 import * as cli from "../definitions/cli";
+import * as signingReleaseHook from "./release-hooks/signing";
 import { AccessKey, Account, App, CollaboratorMap, CollaboratorProperties, Deployment, DeploymentMetrics, Headers, Package, PackageInfo, Session, UpdateMetrics } from "code-push/script/types";
 
 var configFilePath: string = path.join(process.env.LOCALAPPDATA || process.env.HOME, ".code-push.config");
@@ -58,17 +59,17 @@ interface ILoginConnectionInfo {
     noProxy?: boolean; // To suppress the environment proxy setting, like HTTP_PROXY
 }
 
-interface IPackageFile {
-    isTemporary: boolean;
-    path: string;
-}
-
 export interface UpdateMetricsWithTotalActive extends UpdateMetrics {
     totalActive: number;
 }
 
 export interface PackageWithMetrics {
     metrics?: UpdateMetricsWithTotalActive;
+}
+
+interface ReleaseFile {
+    sourceLocation: string;     // The current location of the file on disk
+    targetLocation: string;     // The desired location of the file within the zip
 }
 
 export var log = (message: string | Chalk.ChalkChain): void => console.log(message);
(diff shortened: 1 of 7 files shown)


cli/script/command-executor.ts
```diff
@@ -1091,88 +1092,132 @@ function patch(command: cli.IPatchCommand): Promise<void> {
...
+        signingKeyPath: command.signingKeyPath,
+        type: command.type
+    };
+
+    var hooks: cli.ReleaseHook[] = [ signingReleaseHook.default, coreReleaseHook ];
+
+    var releaseHooksPromise = hooks.reduce((accumulatedPromise: Q.Promise<cli.IReleaseCommand>, hook: cli.ReleaseHook) => {
+        return accumulatedPromise
+            .then((modifiedCommand: cli.IReleaseCommand) => {
+                currentCommand = modifiedCommand || currentCommand;
```
geof90: Could you give some context into what we are trying to do here? Why will the command need to be modified?

cli/script/command-executor.ts
```diff
@@ -1091,88 +1092,132 @@ function patch(command: cli.IPatchCommand): Promise<void> {
...
+        appStoreVersion: command.appStoreVersion,
+        deploymentName: command.deploymentName,
+        description: command.description,
+        disabled: command.disabled,
+        mandatory: command.mandatory,
+        path: command.path,
+        rollout: command.rollout,
+        signingKeyPath: command.signingKeyPath,
+        type: command.type
+    };
```
geof90: If the interface changes, we'd have to update this as well. Could we just do
```
var curentCommand = <cli.IReleaseCommand> Object.assign({}. command);
```

cli/script/command-parser.ts
```diff
@@ -381,12 +381,13 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
...
             .example("release MyApp ./platforms/ios/www 1.0.3 -d Production -r 20", "Releases the \"./platforms/ios/www\" folder and all its contents to the \"MyApp\" app's \"Production\" deployment, targeting the 1.0.3 binary version and rolling out to about 20% of the users")
             .option("deploymentName", { alias: "d", default: "Staging", demand: false, description: "Deployment to release the update to", type: "string" })
             .option("description", { alias: "des", default: null, demand: false, description: "Description of the changes made to the app in this release", type: "string" })
             .option("disabled", { alias: "x", default: false, demand: false, description: "Specifies whether this release should be immediately downloadable", type: "boolean" })
             .option("mandatory", { alias: "m", default: false, demand: false, description: "Specifies whether this release should be considered mandatory", type: "boolean" })
+            .option("signingKey", { alias: "k", default: false, demand: false, description: "Specifies the location of a RSA private key to sign the release with", type: "string" })
```
geof90: nit: I would prefer the alias to be "-s" or "-sign" (the latter same as OpenSSL), since we already use "-k" for listing deployment keys, so that might prevent some confusion

cli/script/release-hooks/signing.ts
```diff
@@ -0,0 +1,87 @@
...
+
+var CURRENT_CLAIM_VERSION: string = "1.0.0";
+var METADATA_FILE_NAME: string = ".codepushrelease"
+
+interface CodeSigningClaims {
+    claimVersion: string;
+    contentHash: string;
+}
+
+export default function sign(command: cli.IReleaseCommand): q.Promise<cli.IReleaseCommand> {
```
geof90: Any reason why you decided to go with `export default` instead of `export = `?

geof90:
Do you think it would be worth adding another hook to delete the generated signature file after the update is released? It's probably minor but I think we don't want to contaminate developers' release contents folders.

geof90:
This is awesome, LGTM!

cli/script/release-hooks/signing.ts
```diff
@@ -0,0 +1,87 @@
...
+
+            return hashUtils.generatePackageHashFromDirectory(command.path, path.join(command.path, ".."));
+        })
+        .then((hash: string) => {
+            var claims: CodeSigningClaims = {
+                claimVersion: CURRENT_CLAIM_VERSION,
+                contentHash: hash
+            };
+
+            return q.nfcall<string>(jwt.sign, claims, signingKey, { algorithm: "RS256" })
```
geof90: Going forward, we should make this algorithm configurable/inferred based on the key.

silhouettes:
Linking this to #307, #233

silhouettes:
Closing this for now but may re-open later when work on this feature is resumed.

dsernst:
@silhouettes: Any updates?

silhouettes:
@dsernst - Yes, @sergey-akhalkov has picked up this feature and is working on it.

Sergey, there are work-in-progress PR's on the Service, CLI and RN SDK that may reduce the work you need to do. Let me know if you'd like me to walk you through it sometime.

sergey-akhalkov:
Hi @silhouettes, got it, thank you!

PR #320, opened 2016-09-17
Keep temp folder around when using release-react command

bsy:
Hey guys, I need to have access to the main.jsbundle (to upload to Sentry.io) after uploading the latest release using the `release-react` command.  The temp directory is already cleared out in `createEmptyTempReleaseFolder`.

Also npm was complaining about the package name and version missing when installing from a git url.


silhouettes:
@bsy I'm closing this because I think #382 will address what you want and also allow you to control the output directory :)

Issue #338, opened 2016-10-24
Release from ManagementSDK - illegal operation on a directory, read


daniel-beard:
- When using the ManagementSDK to release directly to a deployment, I get this error:

```

events.js:154
      throw er; // Unhandled 'error' event
      ^

Error: EISDIR: illegal operation on a directory, read
    at Error (native)
```

Here's the call I'm making (app and deployment names changed):

```
codepush.release( 
'MyApp',
'MyDeployment',
'/tmp/resources/1787/ios',
'1.1.x',
{ description: 'cb53d074b13c8340704bb27243ca557ffda8cb6c 1787'
rollout: 5
});
```

- Any help would be much appreciated!

silhouettes:
Is it possibly a permissions issue on the directory specified here? If you copy that folder to your home directory and `chmod 644` it, will it then work?

daniel-beard:
@silhouettes the permissions are already set to allow all: `drwxrwxrwx` for that folder.

silhouettes:
Ah - my apologies, I think the release function expects the folder to be zipped up, and the path to the zip to be specified. If you zip it up, does that help?

If you're curious, here is what the CLI does on `release`: https://github.com/Microsoft/code-push/blob/master/cli/script/command-executor.ts#L1100
And on `release-react`, which ends up calling `release`: https://github.com/Microsoft/code-push/blob/master/cli/script/command-executor.ts#L1263

This logic could definitely be refactored into the management SDK, currently it's an extremely thin layer.

daniel-beard:
Nice, I'll test this out. Thanks for the prompt reply!

daniel-beard:
Ok, I got it working on my side, but it would be super nice to update the ManagementSDK to be able to take either a path or a .zip file. (As this limitation wasn't clear at all from the documentation).

silhouettes:
Gotcha. Will keep this in mind moving forward, although in the short-term we're unlikely to have the time to implement this ourselves (although community contributions are always welcome!)

silhouettes:
@daniel-beard Thanks so much for your great contribution! This has been published in `code-push@1.11.0-beta`. Closing this now, but feel free to come back if you need anything else!

daniel-beard:
Thanks, this is going to help us a lot. :)
Will definitely be contributing in the future as well.

PR #340, opened 2016-10-26
Allow folder releases from management SDK

daniel-beard:
- Allows releases from folders from the management SDK (i.e. we no longer have to create zip files to release from the SDK)
- Addresses: https://github.com/Microsoft/code-push/issues/338
- Moves the code that switched on file vs folder and produces zip files out of the CLI tool and into the managementSDK.
- @silhouettes


cli/script/command-executor.ts
```diff
@@ -1169,21 +1113,15 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
         isMandatory: command.mandatory,
         rollout: command.rollout
     };
-
-    return getPackageFilePromise
-        .then((file: IPackageFile): Promise<void> => {
+    
+    return Promise<void>((resolve, reject) => {
```
silhouettes: I don't think this promise will ever finish, because `resolve`/`reject` is never called. I think if you just remove this enclosing promise (leaving `return sdk.isAuthenticated()...` below), it'll work perfectly though!

sdk/script/types.ts
```diff
@@ -17,4 +17,9 @@ export interface Session {
     machineName: string;
 }
 
+export interface IPackageFile {
```
silhouettes: This seems to only be used internally within the SDK (after your changes the CLI no longer needs it). I'm wondering if we could just move this interface into `management-sdk.ts` and rename it to something like `PackageFile`? We do have a few legacy naming inconsistencies as you can see :)

sdk/script/management-sdk.ts
```diff
@@ -1,30 +1,23 @@
...
-
-if (typeof window === "undefined") {
-    fs = require("fs");
-} else {
-    fs = {
-        createReadStream: (fileOrPath: string): void => {
-            throw new Error("Tried to call a node fs function from the browser.");
-        }
-    }
-}
```
silhouettes: Okay, it's possible this has gotten out of date. In that case I'm fine with this just being removed, and then we can patch it up later for the browser when/if it's ever needed!

daniel-beard:
@silhouettes updated

PR #343, opened 2016-10-27
Fix failure to unlink temporary file

daniel-beard:
- Need to remove the `packageFile` and not the File object, otherwise the temporary `.zip` file sticks around after a `release` command.
- @silhouettes


silhouettes:
Published in `code-push@1.11.1-beta`

Issue #339, opened 2016-10-25
Can not perform rollback if deployment name is same as CodePush CLI command


ca057:
As stated in the title, I want to perform a rollback of an code-push update. Apparently my deployment name is "release", so it is the same as the code-push command.

**Actual Behaviour**
If I run `code-push rollback <AppName> release`, it does not perform the rollback but instead shows me usage guide for the release command.

**Expected Behaviour**
If I run the command, it interprets "release" as deployment name and not as new command.

My code-push version is `1.12.6-beta`.

max-mironov:
Hello @ca057 . fix for your issue is added as pull request to code-push repo at https://github.com/Microsoft/code-push/pull/360. Also you can try to update directly from my fork repo https://github.com/max-mironov/code-push/tree/update_yargs_to_6.5.0 if needed. Thank you!

sergey-akhalkov:
Closing this due to fix #360 has been merged, please feel free to reopen if you see any issues or have any questions.

PR #360, opened 2016-12-30
Update yargs to 6.5.0

max-mironov:
This should fix issue https://github.com/Microsoft/code-push/issues/339 "Can not perform rollback if deployment name is same as CodePush CLI command #339"


max-mironov:
@silhouettes The issue described in https://github.com/Microsoft/code-push/issues/339 was fixed in yargs 4 and upper versions. I've looked at change logs at https://github.com/yargs/yargs/blob/v6.5.0/CHANGELOG.md#v400-2016214-127-0700 but unfortunately was unable to find the exact notice about the described issue (I think may be fixed in this PR https://github.com/yargs/yargs/pull/368). So I've only determined that versions below 4.0 have the same problem;)

The most part of changes in API were introduced in 4.0 version but I suppose that if it is needed to test this changes anyway it would be better to use the latest version because they might have fixed any other issues as well.
So the API methods were slightly changed. For now `demand() ` points to the number of arguments followed by the command e.g. 
`code-push release <appName> <updateContentsPath> <targetBinaryVersion>` requires 3 non-option arguments followed by the 'release' command, so we changed this to` demand(3, 3)` and etc.
I've tried to do my best to verify that all CLI commands are running as expected with updated yargs. But we can double check it again if needed,

cli/script/command-parser.ts
```diff
@@ -316,7 +316,7 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
         isValidCommandCategory = true;
         isValidCommand = true;
         yargs.usage(USAGE_PREFIX + " login [options]")
-            .demand(/*count*/ 1, /*max*/ 2)  // Require one non-optional and one optional argument.
+            .demand(/*count*/ 0, /*max*/ 0)
```
silhouettes: I think we still want max to be one here. We have an undocumented argument that can be used for testing, eg:

`code-push login http://localhost:3000`

cli/script/command-parser.ts
```diff
@@ -306,7 +306,7 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
         isValidCommandCategory = true;
         isValidCommand = true;
         yargs.usage(USAGE_PREFIX + " link")
-            .demand(/*count*/ 1, /*max*/ 2)  // Require one non-optional and one optional argument.
+            .demand(/*count*/ 0, /*max*/ 0)
```
silhouettes: Same as for login command, I think max should be one because we have an undocumented option used for testing

cli/script/command-parser.ts
```diff
@@ -369,7 +369,7 @@ var argv = yargs.usage(USAGE_PREFIX + " <command>")
         isValidCommandCategory = true;
         isValidCommand = true;
         yargs.usage(USAGE_PREFIX + " register")
-            .demand(/*count*/ 1, /*max*/ 2)  // Require one non-optional and one optional argument.
+            .demand(/*count*/ 0, /*max*/ 0)
```
silhouettes: Same as for login/link commands

cli/script/command-parser.ts
silhouettes: While you're at it, could you please leave a (short) comment explaining that, here and in the other places?

max-mironov:
@silhouettes , thanks for pointing at this, I've pushed required changes. I'd like to perform some little more testing of this and I'm going to merge this changes next week.

PR #373, opened 2017-01-25
cli_readme: improved “App Management” chapter

sergey-akhalkov:
To avoid issues like this one https://github.com/Microsoft/cordova-plugin-code-push/issues/189 “App Management” chapter probably should pay attention on the fact, that “using single app both for iOS and Android may cause installation exceptions due to CodePush update (zip package) produced for iOS may have different content than update produced for Android”.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -143,12 +143,14 @@ Before you can deploy any updates, you need to register an app with the CodePush
 code-push app add <appName>
 ```
 
-If your app targets both iOS and Android, we recommend creating separate apps with CodePush. One for each platform. This way, you can manage and release updates to them separately, which in the long run, tends to make things simpler. The naming convention that most folks use is to suffix the app name with `-iOS` and `-Android`. For example:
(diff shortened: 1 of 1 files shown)


cli/README.md
```diff
@@ -143,13 +143,15 @@ Before you can deploy any updates, you need to register an app with the CodePush
...
 
-If your app targets both iOS and Android, we recommend creating separate apps with CodePush. One for each platform. This way, you can manage and release updates to them separately, which in the long run, tends to make things simpler. The naming convention that most folks use is to suffix the app name with `-iOS` and `-Android`. For example:
+If your app targets both iOS and Android, we highly recommend creating separate apps with CodePush. One for each platform. This way, you can manage and release updates to them separately, which in the long run, tends to make things simpler. The naming convention that most folks use is to suffix the app name with `-iOS` and `-Android`. For example:
 
 ```
 code-push app add MyApp-Android
 code-push app add MyApp-iOS
 ```
 
+*NOTE: Using single app both for iOS and Android may cause installation exceptions due to CodePush update (zip package) produced for iOS may have different content than update produced for Android.*
```
silhouettes: Just touching up the English here -

`Using the same app for iOS and Android may cause installation exceptions because the CodePush update package produced for iOS will have different content from the update produced for Android.`

sergey-akhalkov: I've fixed it, thank you!

Issue #376, opened 2017-01-27
Add `bundleOutput` flag to release-react to save generated bundle into specified location


vladimir-kotikov:
> @benedicthsieh commented on Dec 19, 2016
> We'd like to get the code bundle that is generated by release-react so that we can upload it to Sentry. We see that the bundle is put into a temporary directory, but it seems like that directory is cleaned up once the release is done. Is there a way to either retrieve the bundle afterwards (maybe with a code-push CL command), or preserve the output bundle that's created by release-react?

Copied from https://github.com/Microsoft/react-native-code-push/issues/646

axemclion:
I think we should save both the source map and the bundle at a specified location. This could simply be a flag, something like `--outputDir=`.

silhouettes:
This shouldn't be difficult to do - maybe we could ask one of the vendors to do it if the feature is well-specced?

sergey-akhalkov:
Closing this due to https://github.com/Microsoft/code-push/pull/382 has been merged.

PR #382, opened 2017-02-02
code-push-cli: implement “outputDir” parameter

sergey-akhalkov:
“outputDir” parameter is used to specify the path where bundle and sourcemap should be written.

Relate #376

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -429,6 +429,7 @@ code-push release-react <appName> <platform>
 [--sourcemapOutput <sourcemapOutput>]
 [--targetBinaryVersion <targetBinaryVersion>]
 [--rollout <rolloutPercentage>]
+[--outputDir <outputDir>]
 ```
 
 The `release-react` command is a React Native-specific version of the "vanilla" [`release`](#releasing-app-updates) command, which supports all of the same parameters (e.g. `--mandatory`, `--description`), yet simplifies the process of releasing updates by performing the following additional behavior:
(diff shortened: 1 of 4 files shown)


cli/README.md
```diff
@@ -554,6 +555,12 @@ This specifies the relative path to where the generated JS bundle's sourcemap fi
 
 *NOTE: This parameter can be set using either --sourcemapOutput or -s*
 
+#### Output directory parameter
+
+This specifies the relative path to where the assets, JS bundle and sourcemap files should be written. If left unspecified, assets, JS bundle and sourcemap will copied to /tmp/CodePush folder by default.
```
silhouettes: Suggested modifications:
- will be copied
- to the `/tmp/CodePush` folder
- Could you put back-ticks around the directory path?

sergey-akhalkov:
Fixed it, thanks.

cooperka:
@sergey-akhalkov awesome! Can this also be added to the docs online at https://microsoft.github.io/code-push/docs/cli.html#releasing-updates-react-native? Should I open a new issue for this?

cooperka:
This also seems to clobber the entire directory, it deleted everything previously under `ios` when outputting there. Maybe a warning about this would be nice :P

sergey-akhalkov:
Hi @cooperka, thank you for reaching us. No need to create new issue, there is a script that should sync up docs from github with CodePush site, so I'll ping guys who can help us with this.

>This also seems to clobber the entire directory, it deleted everything previously under ios when outputting there. Maybe a warning about this would be nice

Got it, should we add a `note` to the docs that will pay attention on such behavior?

cooperka:
@sergey-akhalkov a note in the docs would be perfect. Thanks!

sergey-akhalkov:
@cooperka, I've just published https://github.com/Microsoft/code-push/pull/467, please let me know if it looks good to you.

PR #399, opened 2017-02-20
command-executor: check for app and deployment exist before releasing an update

sergey-akhalkov:
This validation helps to save about 1 minute or more in case user has typed wrong app name.
In case app (or deployment) does not exist the error will be shown with the following outputs: 
`App "myapp" does not exist` or `Deployment "mydeployment" does not exist`.
So user have not to wait until code-push update is ready for upload.

Relate https://github.com/Microsoft/react-native-code-push/issues/706

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -1207,9 +1207,6 @@ export var releaseCordova = (command: cli.IReleaseCordovaCommand): Promise<void>
                     throw new Error(`Unable to parse "config.xml" in the CWD. Ensure that the contents of "config.xml" is valid XML.`);
                 });
         })
-        .catch((err: Error) => {
-            throw err;
-        })
         .then((parsedConfig: any) => {
             var config: any = parsedConfig.widget;
 
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -1149,58 +1149,66 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
 }
 
 export var releaseCordova = (command: cli.IReleaseCordovaCommand): Promise<void> => {
-    var platform: string = command.platform.toLowerCase();
-    var projectRoot: string = process.cwd();
-    var platformFolder: string = path.join(projectRoot, "platforms", platform);
-    var platformCordova: string = path.join(platformFolder, "cordova");
-    var outputFolder: string;
-
-    if (platform === "ios") {
-        outputFolder = path.join(platformFolder, "www");
-    } else if (platform === "android") {
-        outputFolder = path.join(platformFolder, "assets", "www");
-    } else {
-        throw new Error("Platform must be either \"ios\" or \"android\".");
-    }
-
-    var cordovaCommand: string = command.build ? "build" : "prepare";
-    var cordovaCLI: string = "cordova";
-
-    // Check whether the Cordova or PhoneGap CLIs are
-    // installed, and if not, fail early
-    try {
-        which.sync(cordovaCLI);
-    } catch (e) {
-        try {
-            cordovaCLI = "phonegap";
-            which.sync(cordovaCLI);
-        } catch (e) {
-            throw new Error(`Unable to ${cordovaCommand} project. Please ensure that either the Cordova or PhoneGap CLI is installed.`);
-        }
-    }
-
-    log(chalk.cyan(`Running "${cordovaCLI} ${cordovaCommand}" command:\n`));
-    try {
-        execSync([cordovaCLI, cordovaCommand, platform, "--verbose"].join(" "), { stdio: "inherit" });
-    } catch (error) {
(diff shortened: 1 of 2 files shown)


cli/script/command-executor.ts
```diff
@@ -1149,58 +1167,64 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
...
-
-    if (platform === "ios") {
-        outputFolder = path.join(platformFolder, "www");
-    } else if (platform === "android") {
-        outputFolder = path.join(platformFolder, "assets", "www");
-    } else {
-        throw new Error("Platform must be either \"ios\" or \"android\".");
-    }
+    var releaseCommand: cli.IReleaseCommand = <any>command;
+    return checkIfAppExists(command.appName)
```
silhouettes: Would it make sense to move this check directly to the `release` command, which is called behind the scenes by `code-push release`, `code-push release-react` and `code-push release-cordova`?

cli/script/command-executor.ts
```diff
@@ -178,6 +178,24 @@ function appList(command: cli.IAppListCommand): Promise<void> {
         });
 }
 
+function checkIfAppExists(appName: string): Promise<App> {
+    log(chalk.cyan(`Checking if app with the name "${appName}" exists...\n`));
+    return Promise<App>((resolve, reject) => {
+        sdk.getApps()
```
silhouettes: You can just return `sdk.getApps() ...` directly instead of wrapping in a promise, as these methods return a promise. Below, you can remove the `if (app)` branch, because if nothing is returned, the promise resolves as success, and instead of `reject(error)`, do `throw new Error(...)` which rejects the promise and matches what we are doing elsewhere.

cli/script/command-executor.ts
```diff
@@ -178,6 +178,24 @@ function appList(command: cli.IAppListCommand): Promise<void> {
         });
 }
 
+function checkIfAppExists(appName: string): Promise<App> {
+    log(chalk.cyan(`Checking if app with the name "${appName}" exists...\n`));
```
silhouettes: I don't think we need this line, as it's taken for granted that we will check the app name (all of the other CLI's effectively do the same but don't log this step unless there is an error).

cli/script/command-executor.ts
silhouettes: You can actually call `sdk.getApp(<name>)` directly, there is no need to do this filtering logic on the client side as it is already being done on the server. And the promise will be rejected for you if it does not exist. Actually, given this, you could maybe remove this function completely, and just call `sdk.getApp(<name>)` within the `release` command, with a comment explaining why

silhouettes: Oh, if you do this I guess you won't be able to return a list of candidate apps on failure, but note that none of the other commands return a list when they fail

silhouettes: Actually thinking about it more, it might make even more sense to call [`sdk.getDeployment(<appName>, <deploymentName>)`](https://github.com/Microsoft/code-push/blob/master/sdk/script/management-sdk.ts#L273), with a comment explaining why, because this way the server will validate both the app and deployment and return the appropriate error message.

sergey-akhalkov:
@silhouettes, I've used [`getDeployment(<appName>, <deploymentName>)`](https://github.com/Microsoft/code-push/blob/master/sdk/script/management-sdk.ts#L273) method to validate for app and deployment exist - thanks for idea! Please verify.

cli/script/command-executor.ts
sergey-akhalkov: I think it's better to put such validation before preparing update contents because it could save time - user have not to wait until `cordova prepare` or  `react-native bundle`  tasks finished.

cli/script/command-executor.ts
```diff
@@ -1149,58 +1149,66 @@ export var release = (command: cli.IReleaseCommand): Promise<void> => {
...
 
-    return configPromise
-        .catch((err: any) => {
-            throw new Error(`Unable to parse "config.xml" in the CWD. Ensure that the contents of "config.xml" is valid XML.`);
+            return configPromise
+                .catch((err: any) => {
+                    throw new Error(`Unable to parse "config.xml" in the CWD. Ensure that the contents of "config.xml" is valid XML.`);
+                });
+        })
+        .catch((err: Error) => {
```
silhouettes: You don't need to add this catch here. A `.catch()` basically catches a rejected promise, so that subsequent `.then()` clauses can execute. However, if you `throw` within a promise callback as you are doing here, the exception is caught by the promise library (`q`) and wrapped in a rejected promise again. So, this is effectively a no-op.

Note that this function (`releaseCordova`) returns a promise, which is the final output after each `.then()` executes. It is okay for the returned promise to be rejected, and in fact we recommend that for error handling. Note that rejected promises should end up [here](https://github.com/Microsoft/code-push/blob/master/cli/script/cli.ts#L15), where the error will be logged.

cli/script/command-executor.ts
silhouettes: Ah - good point!

cli/script/command-executor.ts
sergey-akhalkov: Got it, thanks for the explanation!

PR #402, opened 2017-02-23
Make the prompt defaults to 'no' if no input is entered

max-mironov:
Make the message more severe than it currently for app rm and deployment rm commands


max-mironov:
@silhouettes can you please review? This is in accordance with https://github.com/Microsoft/react-native-code-push/issues/715

silhouettes:
Seeing as Max will be out for a few days, I've made my desired changes directly on the branch. @buptkang - could you please review the final result?

Issue #335, opened 2016-10-14
How have all users update to latest version


rochestb:
Hello,

I have users on a variety of versions and I need them all to be updated to the latest roll out (2.1). From what I can tell when I look at the history is that our of 12k users, none of them have or are able to update the latest version. (Active: 0% (0 of 12,409)   Total: 0 (1 pending) ). Do I need to target binaries and specify which version they should update to? 

`$ code-push release-react myApp ios *`
`$ code-push promote myApp Staging Production`

I'm having a hard time understanding how my rollouts affect users on different versions. 

I am on 1.12.3-beta

rgerd:
Have you tried using the `--mandatory` flag?
https://microsoft.github.io/code-push/docs/cli.html#releasing-updates-general

orlando:
Facing this problem too, If we use a range it doesn't update. But if we target the specific version then the update goes through

patrickomeara:
@orlando I have a similar issue how are you targeting the specific version?

orlando:
@patrickomeara so the issue is not a code-push issue, is how node-semver works. Ranges only works for non pre-release versions. https://github.com/npm/node-semver#prerelease-tags

> For example, the range >1.2.3-alpha.3 would be allowed to match the version 1.2.3-alpha.7, but it would not be satisfied by 3.4.5-alpha.9, even though 3.4.5-alpha.9 is technically "greater than" 1.2.3-alpha.3 according to the SemVer sort rules. The version range only accepts prerelease tags on the 1.2.3 version. The version 3.4.5 would satisfy the range, because it does not have a prerelease flag, and 3.4.5 is greater than 1.2.3-alpha.7.
>
>The purpose for this behavior is twofold. First, prerelease versions frequently are updated very quickly, and contain many breaking changes that are (by the author's design) not yet fit for public consumption. Therefore, by default, they are excluded from range matching semantics.

So what I did is removing our prerelease tags from our staging environment and just using the same version syntax as production

sergey-akhalkov:
@orlando, @patrickomeara, please let me know if I could help of if the issue was resolved.

orlando:
@sergey-akhalkov no problem here, is the expected behaviour of node-semver. What we need to do maybe is to put a reference in the code-push documentation

sergey-akhalkov:
@orlando, got it, thank you, I'll update docs to make this more clear.

sergey-akhalkov:
Closing this due to #405 has been merged.

PR #405, opened 2017-02-27
cli-readme: ranges only works for non pre-release versions

sergey-akhalkov:
Due to https://github.com/npm/node-semver#prerelease-tags ranges only works for non pre-release versions.
If you want to update a version with pre-release tags, then you need to write the exact version you want to update (`1.12.3-beta` for example)

Relates to https://github.com/Microsoft/code-push/issues/335


sergey-akhalkov:
Fixed it, thanks!

Issue #392, opened 2017-02-09
Production release, runs the command cordova build


srameshr:
```
code-push release-cordova my_app ios -t 1.2.0  --description "Test" --build  deploymentName Production 
```

This command runs ```cordova build ios``` and not ```cordova build ios --release```

max-mironov:
Hello @srameshr, we have implemented this in https://github.com/Microsoft/code-push/pull/409.
Closing this for now, please let us know if you see any issues or have any questions.

gs-sramesh:
@max-mironov Thanks!

PR #409, opened 2017-03-06
Add  build type option to release cordova command

max-mironov:
See https://github.com/Microsoft/code-push/issues/392 for details

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -147,7 +147,7 @@ Before you can deploy any updates, you need to register an app with the CodePush
 code-push app add <appName>
 ```
 
-If your app targets both iOS and Android, we highly recommend creating separate apps with CodePush. One for each platform. This way, you can manage and release updates to them separately, which in the long run, tends to make things simpler. The naming convention that most folks use is to suffix the app name with `-iOS` and `-Android`. For example:
+If your app targets both iOS and Android, please *create separate apps* with CodePush (see NOTE below for details). One for each platform. This way, you can manage and release updates to them separately, which in the long run, tends to make things simpler. The naming convention that most folks use is to suffix the app name with `-iOS` and `-Android`. For example:
 
 ```
 code-push app add MyApp-Android
--- a/cli/README.md
+++ b/cli/README.md
@@ -147,7 +147,7 @@ Before you can deploy any updates, you need to register an app with the CodePush
 code-push app add <appName>
 ```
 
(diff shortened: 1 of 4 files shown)


cli/README.md
```diff
@@ -580,6 +580,7 @@ code-push release-cordova <appName> <platform>
 [--targetBinaryVersion <targetBinaryVersion>]
 [--rollout <rolloutPercentage>]
 [--build]
+[--isReleaseBuildType]
```
silhouettes: Both the `--build` and `--isReleaseBuildType` flags need to be alphabetically ordered above

cli/README.md
```diff
@@ -643,6 +644,10 @@ This is the same parameter as the one described in the [above section](#disabled
 
 Specifies whether you want to run `cordova build` instead of `cordova prepare` (which is the default behavior), when generating your updated web assets. This is valuable if your project includes before and/or after build hooks (e.g. to transpile TypeScript), and therefore, having CodePush simply run `cordova prepare` isn't sufficient to create and release an update. If left unspecified, this defaults to `false`.
 
+#### IsReleaseBuildType parameter
```
silhouettes: Both the `Build` and `IsReleaseBuildType` sections need to be alphabetically ordered above

cli/script/command-executor.ts
```diff
@@ -1175,7 +1175,9 @@ export var releaseCordova = (command: cli.IReleaseCordovaCommand): Promise<void>
                 throw new Error("Platform must be either \"ios\" or \"android\".");
             }
 
-            var cordovaCommand: string = command.build ? "build" : "prepare";
+            var cordovaCommand: string = command.build ? 
+                (command.isReleaseBuildType ? "build --release" : "build") :
```
silhouettes: We normally name the flags that we pass through to the underlying `react-native` or `cordova` CLI's the same as they would be in the original CLI. This generally makes them more discoverable/consistent for people who already have been using the underlying CLI. I understand why it might have made sense to name it otherwise, but with all things considered do you think it makes sense to just call the flag `release` instead of `isReleaseBuildType`? (I'll leave this up to you).

max-mironov: @silhouettes thank you for your feedback. I thought initially about this and I was a little bit confused about naming it `release `because it would be semantically closer to something related to our `code-push release` command which actually perform the release of our updates to code push server and has nothing in common with release build flag. So I think naming this flag as  `release` may lead to some confusion. If you don't mind I'd prefer to avoid 'release' name here in this particular case.

silhouettes: Okay sure - just wanted to make sure that was considered, but sounds like you have :)

Issue #412, opened 2017-03-09
code-push patch didn't work


laowenlu:
I use 
`code-push patch HybridDemo Debug -l v13 --des "close v13" --disabled true --mandatory false`;
after
`code-push deployment history HybridDemo Debug --format json  -a`
then log
```
...
{
    "description": "close v13",
    "isDisabled": false,
    "isMandatory": true,
    "rollout": null,
    "appVersion": "1.0.0",
    "packageHash": "92f5184adef14fae6f856e9b84f9bfef179cfa32f79c0b46b15e453dff1f0edd",
    "blobUrl": "https://codepush.blob.core.windows.net/storagev2/0IawVio3OtfUNhKy5azUxgibXAZTN1_HEN05-",
    "size": 474487,
    "manifestBlobUrl": "https://codepush.blob.core.windows.net/storagev2/kpxsdQ-thhl_uF1dV-s5jpksIVZcN1_HEN05-",
    "releaseMethod": "Upload",
    "uploadTime": 1480649820843,
    "label": "v13",
...
```

There `isDisabled` and `isMandatory` is no change

max-mironov:
Hello @laowenlu thank you so much for pointing to the issue, Fix for it has been merged into the master branch. 
I'm closing the issue for now, please let us know if you still see any issues after updating from master branch of code-push.

PR #413, opened 2017-03-09
Fixed issue with getting bool values for yargs 6.5.0

max-mironov:
Should fix issue https://github.com/Microsoft/code-push/issues/412


max-mironov:
Verified the above behavior and it is correct!

PR #417, opened 2017-03-14
Small tweak to fix unit tests

silhouettes:
(Add ?w=1 to the end of the URL in the file view tab to ignore whitespace changes)

The unit tests pass `command.label` as `undefined` instead of `null`, which means the output of the `promote` command will be wrong in the unit tests.

@max-mironov, @sergey-akhalkov - this is not a customer-facing issue so it's not a big deal at all, and it's my fault for not mentioning it, but please make sure you run `gulp test` before you submit PR's, and maybe even add test cases if you can. This applies to the CLI and service repos only, as the test framework in the SDK repos is incomplete. Also, please review this change, and I can merge it after.

### DIFF
--- a/cli/script/command-executor.ts
+++ b/cli/script/command-executor.ts
@@ -94,8 +94,8 @@ export var confirm = (message: string = "Are you sure?"): Promise<boolean> => {
                 resolve(true);
             } else {
                 if (!rejected){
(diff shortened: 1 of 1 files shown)


max-mironov:
@silhouettes got it, will always run `gulp test` before PR's for this repositories. 
Your PR looks good!

Issue #419, opened 2017-03-16
Warnings from yarn regarding node-uuid


fungilation:
On every `yarn upgrade`:
```
warning react-native-code-push > code-push > node-uuid@1.4.7: use uuid module instead
warning react-native-code-push > xcode > node-uuid@1.4.7: use uuid module instead
```

Perhaps the warning can be fixed as suggested, if it's "correct"?

Using `code-push` 1.8.0-beta

max-mironov:
Looks like the package `node-uuid` package on npm has been deprecated but the repo hasn't been because its now serving up the source code for `uuid`. More details here https://github.com/kelektiv/node-uuid/issues/155

@silhouettes - I've searched through code-push project and it seems that we are not using `node-uuid` module any more at this time. What do you think, is it ok if we remove this dependency completely from our code-push `package.json`?

silhouettes:
@max-mironov great catch! yes, please remove this dependency completely.

silhouettes:
Oh, @max-mironov - the xcode dependency also uses node-uuid, so the warning will still show up even after my PR is merged. You might need to look at that separately.

max-mironov:
@silhouettes - I've looked closer at this and it turns out that nothing we can do with this warning. 
As it is stated here  https://github.com/react-community/create-react-native-app/issues/86 this warning appears due to our `code-push-react-native` (as well as `react-native`) have dependency upon `xcode `package which is in its turn dependent upon `node-uuid` package. 
So I believe nothing we can do here until `xcode` package will update their dependencies.

silhouettes:
@max-mironov - gotcha! Thanks for looking into that. So I guess we've done all we can on our end.

max-mironov:
@fungilation thank you for referencing the issue in `node-xcode` repository. Currently we'll left this item open and wait for new release of `xcode` to bump dependency accordingly.

fungilation:
it's version bumped to include the move to uuid: https://github.com/alunny/node-xcode/pull/117#issuecomment-290856044

max-mironov:
Hey @fungilation, thanks for heads up!
We have updated xcode dependency on our side [in this PR](https://github.com/Microsoft/react-native-code-push/pull/776).

However currently we will still see this warnings for 2 reasons:
1) `react-native` package depends upon old xcode package, we cannot force them to update it immediately
[...]

PR #422, opened 2017-03-23
Remove redundant dependencies

silhouettes:
In addition to the redundancy @max-mironov found in #419, I discovered some more as part of my cryptography audit.

@max-mironov - could you review this one?


max-mironov:
@silhouettes - looks good! Also walked through other dependencies in `code-push` and verified that we have no more unused packages.

Issue #377, opened 2017-01-27
Plain integer version number should be accepted by CodePush.


benjaminb10:
My app version number is 1
I prefer 1 to 1.0
My next version will be number 2, then 3, etc.
I prefer 1, 2, 3 to 1.0, 1.1, 1.2, etc.
_It's here a personal choice and off topic. Just Like what Messenger app from Facebook does, they are at their version number 100 (plain integer 100 and not 100.0)_

But when I put 1 for the value of CFBundleShortVersionString in my Info.plist file,
I have this CodePush error: 

> [Error]  The "CFBundleShortVersionString" key in the "ios/MyApp/Info.plist" file needs to specify a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).

Why does CodePush block a plain integer version number?
It should allow me to use 1 as a version number.

max-mironov:
Closing this as PR #429 was merged to Master.

PR #429, opened 2017-04-12
Plain integer number

max-mironov:
Add ability to use plain integer number for app version

See https://github.com/Microsoft/code-push/issues/377 for details


max-mironov:
Merging this to Master due to the corresponding changes to support plain integer number (adding "0.0" to it, so it means that e.g. v. 2 would be treated as v. 2.0.0 this way) were deployed to code-push service side.

PR #450, opened 2017-05-26
update cli README-cn

stonexer:


sergey-akhalkov:
Hi @stonexer, thanks for the contribution and sorry for the delayed reply. LGTM.

PR #474, opened 2017-06-12
Read me fix

iageoghe:
Documentation for the addApp method needed updating.


iageoghe:
Updating documentation.

Issue #473, opened 2017-06-12
Outdated docs about updating access-key


PawelWesolowski:
From the [docs](https://github.com/Microsoft/code-push/tree/master/cli#access-keys):

> Finally, if at any point you need to change a key's name and/or expiration date, you can use the following command:

`code-push access-key patch <accessKeyName> --name "new name" --ttl 10d`

When I run it I got:

```
[Error]  Updating access keys is no longer supported.
Please upgrade to the latest version of the CodePush CLI by running 'npm install -g code-push-cli@latest'.
```

After update (now I have cli version 2.0.1-beta) the error is not there but I am getting help about usage:

Usage: code-push access-key <command>

```
Commands:
  add     Create a new access key associated with your account
  remove  Remove an existing access key
  rm      Remove an existing access key
  list    List the access keys associated with your account
  ls      List the access keys associated with your account

Options:
  -v, --version  Show version number  [boolean]
```

There is nothing about -ttl anymore. Does it mean that ttl is unlimited now?

BretJohnson:
@PawelWesolowski Thanks for the feedback here. Yes, in Mobile Center access keys (which Mobile Center calls API Token elsewhere in the UI, like under user settings in the web UI), never expire currently. Mobile Center might add support for expiring tokens in the future, but it's not there currently. Also the tokens currently can't be renamed once created, in Mobile Center. So `patch` went away.

We'll update the doc. I'll keep this issue open until we do. Thanks again for the report.

iageoghe:
Documentation patched.

D1no:
Just fell into that trap trying to set a high expiration for our CI Service. Still wrong here:
https://microsoft.github.io/code-push/docs/cli.html#link-2

iageoghe:
@D1no - I am so sorry to have missed that; documentation has been corrected. Thanks for pointing out the error.

PR #476, opened 2017-06-13
Removed reference to patch for acccess keys

iageoghe:
See https://github.com/Microsoft/code-push/issues/473


iageoghe:
Nice catch Bret; thanks.

PR #492, opened 2017-08-02
Add self-sign bundles support

ruslan-bikkinin:
Add basic support for self-sign bundles. You can specify `-k` option for `release` and `release-react` commands with following path to private RSA key for signing. CLI will sign bundle and produce signed JWT named `.codepushrelease` distributed within update.

Tests and docs update will be appeared in future PRs.

Relates to https://github.com/Microsoft/code-push/issues/307


pniko:
Excellent @ruslan-bikkinin !

PR #516, opened 2017-09-28
Grammar error fix "is order to" -> "in order to"

junjchen:


max-mironov:
@junjchen thanks for fixing!

PR #517, opened 2017-09-28
Update gradle-to-js

FredyC:
Important bugfix there ... https://github.com/ninetwozero/gradle-to-js/pull/9

There is no changelog, but based on list of commits, there is not much of changes from the currently used 0.2.5

Edit: On another look seeing the last comment there, something might be broken. Consider this PR as a starting point then :)

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -30,7 +30,7 @@
     "cli-table": "^0.3.1",
     "code-push": "2.0.2-beta",
     "email-validator": "^1.0.3",
-    "gradle-to-js": "0.2.5",
+    "gradle-to-js": "^1.0.1",
     "jsonwebtoken": "^7.4.1",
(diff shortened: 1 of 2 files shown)


ruslan-bikkinin:
LGTM

ruslan-bikkinin:
But... 
@FredyC i think we don't need those anymore: https://github.com/Microsoft/code-push/blob/master/cli/script/command-executor.ts#L978-L989

FredyC:
@ruslan-bikkinin Thanks for sorting this out quickly. Point of that upstream PR was this error https://github.com/ninetwozero/gradle-to-js/issues/8. Not a big deal as it can be fixed by little edit to gradle file, but is kinda annoying when encountered :)

ruslan-bikkinin:
@FredyC you are welcome)

PR #524, opened 2017-10-12
add privateKey support for release-cordova

afoo:
In preparation for code-signing support for cordova, the code-push cli should support the -k option for the `release-cordova` command as well.

### DIFF
--- a/cli/script/release-hooks/signing.ts
+++ b/cli/script/release-hooks/signing.ts
@@ -17,22 +17,41 @@ interface CodeSigningClaims {
     contentHash: string;
 }
 
+var deletePreviousSignatureIfExists = (command: cli.IReleaseCommand): q.Promise<any> => {
+    var signatureFilePath = path.join(command.package, METADATA_FILE_NAME);
+    var prevSignatureExists = true;
+    try {
+        fs.accessSync(signatureFilePath, fs.R_OK);
+    } catch (err) {
+        if (err.code === "ENOENT") {
+            prevSignatureExists = false;
+        } else {
+            return q.reject(new Error(
+                `Could not delete previous release signature at ${signatureFilePath}.
+                Please, check your access rights.`)
+            );
+        }
+    }
+
+    if (prevSignatureExists) {
+        console.log(`Deleting previous release signature at ${signatureFilePath}`);
+        rimraf.sync(signatureFilePath);
+    }
+
+    return q.resolve<cli.IReleaseCommand>(command);
+}
+
 var sign: cli.ReleaseHook = (currentCommand: cli.IReleaseCommand, originalCommand: cli.IReleaseCommand, sdk: AccountManager): q.Promise<cli.IReleaseCommand> => {
+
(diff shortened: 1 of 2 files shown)


ruslan-bikkinin:
Hi, @afoo and thank you for submitting this PR! I've been updated it with small improvement and will be updating it in the future because we will need to update docs as well. Thanks again for time you spent working on code signing feature implementation for Cordova SDK, Codepush team are really appreciates your help!

cli/script/release-hooks/signing.ts
```diff
@@ -19,6 +19,14 @@ interface CodeSigningClaims {
 
 var sign: cli.ReleaseHook = (currentCommand: cli.IReleaseCommand, originalCommand: cli.IReleaseCommand, sdk: AccountManager): q.Promise<cli.IReleaseCommand> => {
     if (!currentCommand.privateKeyPath) {
+        if (fs.lstatSync(currentCommand.package).isDirectory()) {
+            // If new update wasn't signed, but signature file for some reason still appears in the package directory - delete it
+            let signatureFilePath = path.join(currentCommand.package, METADATA_FILE_NAME);
+            if (fs.existsSync(signatureFilePath)) {
+                fs.unlinkSync(signatureFilePath);
```
max-mironov: Don't we need here to check for access rights here to delete file? Similar as this is done below for removing the signature? If yes - does it make sense to use utility function for this?

ruslan-bikkinin: @max-mironov I think we don't need to do that. If something wrong with the access rights EPERM will be thrown and command execution will be stopped.

max-mironov: @ruslan-bikkinin 
Probably it could be a good idea to show human friendly message like e.g. this one: https://github.com/Microsoft/code-push/blob/master/cli/script/release-hooks/signing.ts#L57 so users will be aware what's wrong here rather than throwing `EPERM ` exception. What do you think about this?

ruslan-bikkinin: @max-mironov Well, may be. I can do it.

sergey-akhalkov: @ruslan-bikkinin, why don't you just wrapped up the following code into the method and used it for deleting `METADATA_FILE_NAME`? I don't think if it's a good idea to use different versions of the code/implementations to achieve the same goals.
```js
             var prevSignatureExists = true;
             try {
                 fs.accessSync(signatureFilePath, fs.F_OK);
             } catch (err) {
                 if (err.code === "ENOENT") {
                     prevSignatureExists = false;
                 } else {
                     return q.reject<string>(new Error(
                         `Could not delete previous release signature at ${signatureFilePath}.
                         Please, check your access rights.`)
                     );
                 }
             }
 
             if (prevSignatureExists) {
                 console.log(`Deleting previous release signature at ${signatureFilePath}`);
                 rimraf.sync(signatureFilePath);
             }
```

PR #533, opened 2017-11-09
Update return type of management sdk release method

belemaire:
This PR updates the return type of the management sdk `release` method, to return a `Promise<Package>` instead of a `Promise<void>`.

To conduct some code push automation part of a project, we need to get back the `label` that was created following a deployment through `release` method code. Unfortunately, because the `release` method is returning a `Promise<void>` we don't get this information back, even though it is contained within the successful server response.

Looking at the response payload we get back from the server following a successful deployment through `release` method call,

```json
{
    "package": {
        "isMandatory": true,
        "appVersion": "1.0",
        "packageHash": "3fc3828e674d8903fdc7c0dd4924938946f6b76314e1317cdef10fa946b89060",
        "blobUrl": "https://codepush.blob.core.windows.net/storagev2/wb2Ee7_OpeMx67tr2hxoeJRhYaZe4a9b7d8f-80d4-459f-b625-566c1ae80af1",
        "size": 145160,
        "releaseMethod": "Upload",
        "uploadTime": 1510263394211,
        "label": "v4",
        "releasedBy": "lemaireb@gmail.com",
        "rollout": 100
    }
}
```

I figured out that the `package` object returned was matching the [Package type]( https://github.com/Microsoft/code-push/blob/19c6f2442e31c73ebb499b07cb1f903c151bacdd/definitions/rest-definitions.d.ts#L147)

It makes sense to return this object instead of `void`, as it contains the `label` along with other interesting information that can be used/stored if needed (or even logged to the user in some way when releasing through CLI command).

### DIFF
--- a/sdk/script/management-sdk.ts
+++ b/sdk/script/management-sdk.ts
@@ -331,19 +331,17 @@ class AccountManager {
                             return;
                         }
 
-                        if (res.ok) {
-                            resolve(<Package>(JSON.parse(res.text).package));
-                        } else {
-                            try {
-                                var body = JSON.parse(res.text);
-                            } catch (err) {
-                            }
+                        try {
+                            var body = JSON.parse(res.text);
+                        } catch (err) {
+                            reject(<CodePushError>{ message: `Could not parse response: ${res.text}`, statusCode: AccountManager.ERROR_INTERNAL_SERVER });
+                            return;
+                        }
 
-                            if (body) {
-                                reject(<CodePushError>{ message: body.message, statusCode: res && res.status });
-                            } else {
-                                reject(<CodePushError>{ message: res.text, statusCode: res && res.status });
-                            }
+                        if (res.ok) {
+                            resolve(<Package>body.package);
+                        } else {
+                            reject(<CodePushError>{ message: body.message, statusCode: res && res.status });
                         }
                     });
             });
--- a/sdk/script/management-sdk.ts
+++ b/sdk/script/management-sdk.ts
@@ -299,9 +299,9 @@ class AccountManager {
             .then((res: JsonResponse) => res.body.history);
     }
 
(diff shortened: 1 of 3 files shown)


sdk/script/management-sdk.ts
```diff
@@ -332,7 +332,7 @@ class AccountManager {
                         }
 
                         if (res.ok) {
-                            resolve(<void>null);
+                            resolve(<Package>(JSON.parse(res.text).package));
```
sergey-akhalkov: I believe we need to wrap up `JSON.parse` into the `try/catch`, just in case. (please see the same approach below https://github.com/Microsoft/code-push/pull/533/files#diff-84f033a739bf58615844d4686a717544R337)

belemaire:
@sergey-akhalkov Thanks !

Please check latest commit addressing your comment. 
Just extracted the response deserialization part (`JSON.parse`) as it is now needed for both success and error response.

In case of a deserialization failure, independently of the http`res` code value, just `reject` with the response text as the `CodePushError` message.

If you don't need any further changes, please let me know, I can squash if needed.

belemaire:
@max-mironov 
Thanks for your review.
Yes updated at the caller site (found only one) as well as in documentation of the Management SDK.
I have not updated caller behavior though. As mentioned in my original commit, CLI release command log message could be enhanced with more info if needed.

max-mironov:
@belemaire got it, thanks. So merging this, hope we can release it by the end of this week.

belemaire:
Thanks @max-mironov , looking forward to the release.

max-mironov:
@belemaire we've released new versions of CLI & SDK (https://github.com/Microsoft/code-push/releases/tag/v2.1.3-beta), thanks again for contribution!

PR #537, opened 2017-11-19
Update error message to use `--target-binary-version`.

joshuapinter:
Instead of the old `--targetBinaryVersion`.


joshuapinter:
Nevermind, this was meant for `appcenter`. Just upgrade so getting confused.

PR #538, opened 2017-11-21
Update return type of SDK promote method 

belemaire:
In continuation to my previous PR https://github.com/Microsoft/code-push/pull/533, this PR updates the return type of the SDK `promote` method to return the `Package` object part of the server response instead of returning `void`.

Part of an ongoing project, calling in the CodePush SDK, we need the data resulting from the promotion.

It seems that the server call associated to `promote` will always return a body with a package object resulting from the promotion. Even the test was mocking a response with a package, so it seems that the server will always return a package response body in case of success.
If not (I'm doubting it), what could be done is changing the return type to `Promise<Package|void>` and changing the `expectResponseBody` back to false, and return the response Package if present, otherwise don't return anything, and let the caller act accordingly.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -69,7 +69,7 @@ The `code-push` module exports a single class (typically referred to as `CodePus
 
 - __patchRelease(appName: string, deploymentName: string, label: string, updateMetadata: PackageInfo): Promise&lt;void&gt;__ - Updates the specified release's metadata with the given information.
 
-- __promote(appName: string, sourceDeploymentName: string, destinationDeploymentName: string, updateMetadata: PackageInfo): Promise&lt;void&gt;__ - Promotes the latest release from one deployment to another for the specified app and updates the release with the given metadata.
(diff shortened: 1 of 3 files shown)


belemaire:
:+1:

belemaire:
@max-mironov : Are you planning on publishing a new version of the SDK including this change soon ? We're waiting for this to release a new version of our platform. Thanks !

max-mironov:
@belemaire sure, let's do it in the nearest days!

belemaire:
Thanks @max-mironov ! 👍

max-mironov:
@belemaire we have released new verison - https://github.com/Microsoft/code-push/releases/tag/v2.1.4. Thank you!

belemaire:
Thanks @max-mironov for the quick release, greatly appreciated, have a good week :)

belemaire:
@max-mironov FYI, don't know if that is intended, but there is a mismatch between the published version and the tag.
`2.0.4` is the latest published version while `2.1.4` is the tag.
That's OK, I noticed it and picked `2.0.4` but it might seem confusing to other people.
It seems like it's not the first version for which this is the case where minor digit is different.

belemaire:
Oh sorry, my bad, just realized it was the `CLI` version that was tracked by tags, not the `SDK` version. CLI is what matters to most users so I guess it's OK ;)

Issue #544, opened 2017-12-09
Updates released with --outputDir option loses codesign signature


PaperMonster:
Hi, thanks for a great library which was one of the main reasons my company switched to React Native.

I realized from deployment history that recent updates haven't been installed nor downloaded, and code-signed iOS app log contains following error message:
````
Error! Public key was provided but there is no JWT signature within app bundle to verify
````
After some investigation, we found that the problem only occurs when the update was released with option `--outputDir` AND `--privateKeyPath`. For example:
````
code-push release-react myapp-ios ios --outputDir build --privateKeyPath ~/myapp-codepush-private.pem
````
I looked into `build` folder where the output were generated and file `.codepushrelease` file was there. It seems the update had lost this file somehow when it reached the client.
If I omit `--outputDir build`, the client receives and apply the update successfully. 
We need `--outputDir` option for crash reporting service. Currently we generate the output manually, but it would be nice being able to do it along with CodePush release.

CodePush CLI version: 2.1.4
react-native-code-push version: 5.2.1

sergey-akhalkov:
Hi @PaperMonster, thanks for reporting the issue! I'm going to investigate it and let you know if I have any results or questions, thanks for the patience.

sergey-akhalkov:
Hi @PaperMonster, could you please take a look at https://github.com/Microsoft/code-push/pull/545 and let me know if it works for you or not? Also let me know if you see any issues or have any questions.

sergey-akhalkov:
@PaperMonster, to test the fix you can follow these instructions:
1. download changs from the #545 
2. `cd code-push`
3. `gulp build`
4. `cd cli/bin`
5. `npm install`
6. `code-push release-react myapp-ios ios --outputDir build --privateKeyPath ~/myapp-codepush-private.pem`

Please let me know if you have any questions or see any issues.

PaperMonster:
@sergey-akhalkov The fix works beautifully. Thank you for your hard work. 👍

sergey-akhalkov:
@PaperMonster, thanks, good to know it works, going to publish an release at the end of this or in the beginning of the next week.

max-mironov:
Hey @PaperMonster we have release new version of code push https://github.com/Microsoft/code-push/releases/tag/v2.1.5
Please let us know if you see any issues.

PR #545, opened 2017-12-12
fix issue w/ outputDir in case of using code-signing

sergey-akhalkov:
relates to https://github.com/Microsoft/code-push/issues/544

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -154,7 +154,7 @@
         "slash": "1.0.0",
         "superagent": "1.8.5",
         "superagent-proxy": "1.0.2",
-        "yazl": "2.4.2"
+        "yazl": "2.4.3"
       }
     },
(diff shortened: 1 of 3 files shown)


sergey-akhalkov:
@max-mironov, @ruslan-bikkinin JFYI - going to fix the same in the https://github.com/Microsoft/appcenter-cli after approval from @PaperMonster that the issue is fixed w/ this PR

shashankwadi:
Is this Fixed? Not working for mw still RN version is .51

SudoPlz:
Hey @sergey-akhalkov @max-mironov , thanks for all the great work.
Does this PR mean that the path will always be `<ourBuildOutputPath>/CodePush/`

Looks like the `CodePush/` dir is always created and appended to our build output path.

Thanks

PR #550, opened 2017-12-27
[Acquisition SDK] Add status.code check for zero value

ruslan-bikkinin:
This PR adds additional error information for the case when zero status code was returned during the request to updateCheck server method.
Related to https://github.com/Microsoft/cordova-plugin-code-push/issues/328

### DIFF
--- a/sdk/script/acquisition-sdk.ts
+++ b/sdk/script/acquisition-sdk.ts
@@ -103,10 +103,15 @@ export class AcquisitionManager {
             }
 
             if (response.statusCode !== 200) {
-                callback(new Error(response.statusCode + ": " + response.body), /*remotePackage=*/ null);
+                let errorMessage;
+                if (response.statusCode === 0){
+                    errorMessage = `Couldn't send request to ${requestUrl}. One of the possible reasons for that might be connection problems. Please, check your internet connection.`;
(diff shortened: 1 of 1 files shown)


sdk/script/acquisition-sdk.ts
```diff
@@ -104,8 +104,8 @@ export class AcquisitionManager {
 
             if (response.statusCode !== 200) {
                 let errorMessage;
-                if (response.statusCode === 0){
-                    errorMessage = `Couldn't send request to ${requestUrl}. One of the possible reasons for that might be connection problems. Please, check your internet connection.`;
+                if (response.statusCode === 0) {
+                    errorMessage = `Couldn't send request to ${requestUrl}, xhr.statusCode = 0 was returned. One of the possible reasons for that might be connection problems. Please, check your internet connection. Also, see https://stackoverflow.com/a/26451773 for more details`;
```
sergey-akhalkov: Don't sure if it's a good idea to provide 3rd party URLs in our logs, would be better to provide a link to CodePush issue/comment on github if we really need it.

ruslan-bikkinin: Got it, I removed it.

PR #591, opened 2018-09-27
add cli option for project dir

kousun12:
The motivation for this is that sometimes you want to be able to run the `release-react` command from some other directory (deploying in ci, etc) and it would be nice if you could just pass a cli option in that tells `codepush` where to find the react-native cli, rather than assuming the process is always run from that directory.

### DIFF
--- a/cli/definitions/cli.ts
+++ b/cli/definitions/cli.ts
@@ -207,6 +207,7 @@ export interface IReleaseReactCommand extends IReleaseBaseCommand {
     plistFilePrefix?: string;
     sourcemapOutput?: string;
     outputDir?: string;
+    projectDir?: string;
     config?: string;
 }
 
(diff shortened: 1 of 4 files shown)


yuri-kulikov:
Hi @kousun12,
It's a pretty nice feature, thanks for contribution!
It needs to be tested on our side before we can merge it but the code looks good and it's great you've added the test.
Also we need to update documentation for this changes here: **[/cli/README.md](https://github.com/Microsoft/code-push/blob/master/cli/README.md)**. Could you please push additional commit to update docs or we can take care if ourselves it needed.

NickToropov:
Hey @kousun12! 
We haven’t heard from you in a while, so I am going to go ahead and close this PR. If you have any updates,  reopen it.

PR #593, opened 2018-10-08
Fix button break line on Firefox

diegowifi:
The "Visit App Center" button had a break line on Firefox

screen shot 2018-10-07 at 9 00 23 pm


alexandergoncharov:
Hi @diegowifi,
Thanks for reporting!

I reproduced your issue and confirm that this changes fix it without affecting other browsers.

PR #645, opened 2019-08-14
Update acquisition sdk to work with new service

alexandergoncharov:
* Update types of `UpdateCheckRequest`, `UpdateCheckResponse`, `DeploymentStatusReport`, `DownloadReport`.
* Update logic and test with new names.
* Small code style refactors.

### DIFF
--- a/definitions/rest-definitions.d.ts
+++ b/definitions/rest-definitions.d.ts
@@ -60,19 +60,18 @@ declare module "rest-definitions" {
     }
 
     /*out*/
-    export interface UpdateCheckResponse extends PackageInfo {
-        downloadURL?: string;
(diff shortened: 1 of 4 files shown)


cjonsmith:
Can we get a bump to the next major version since this will be a breaking change? When it comes time to publish, we should also make it a beta release until we have the new RN SDK and Cordova plugin updated.

Otherwise, LGTM 👍

alexandergoncharov:
@cjonsmith , usually we bump version right away before release. But I'll bump version right now for this changes. I'm not sure if we really need to have beta release as we can update CodePush sdk and react-native/cordova plugins at the same time.
So, I'm going to merge this PR and bump version right away before release.

PR #676, opened 2019-11-19
Fix return undefined

victornovy:
Fix return undefined getDeployment. I think it has to change the api test's to return the correct object.


alexandergoncharov:
Hi @victornovy,
Thanks for contributing!

Unfortunately, it was a server-side issue. We already fix it.
I'm going to close this PR for now. Please feel free to reopen it if you have any questions or face this issue again.

PR #730, opened 2020-07-28
Change SERVER_URL

Krasavinigor:
**Description**: Implement the same approach as `appcenter-cli`

### DIFF
--- a/src/script/management-sdk.ts
+++ b/src/script/management-sdk.ts
@@ -42,8 +42,7 @@ class AccountManager {
         OWNER: "Owner",
         COLLABORATOR: "Collaborator"
     };
-    public static SERVER_URL = "https://codepush.appcenter.ms/v0.1/legacy";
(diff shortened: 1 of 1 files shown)


src/script/management-sdk.ts
```diff
@@ -510,9 +509,7 @@ class AccountManager {
             }
         }
 
-        request.set("Accept", `application/vnd.code-push.v${AccountManager.API_VERSION}+json`);
-        request.set("Authorization", `Bearer ${this._accessKey}`);
-        request.set("X-CodePush-SDK-Version", packageJson.version);
```
andreidubov, line 515: Do we still need `packageJson` in this file?

andreidubov: Same about `API_VERSION`.

Krasavinigor: Good catch! Thank you!

Issue #806, opened 2023-01-09
CodePushHttpError pass response as error cause 


retyui:
Hi team! Could you please pass a network response as `Error.prototype.cause`, too simply logic with handling response statuses

Example: 
```tsx
const err = new CodePushHttpError(msg, {cause: response})

if(err.cause.status === 500){
  // my logic
}
```

meta: 
- `code-push@4.0.4`
- [`Error.prototype.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)

davydspotnana:
no

retyui:
.

PR #807, opened 2023-01-10
[Enhancement] Use a network `Response` as `error.cause` #806

retyui:
- Fix for https://github.com/microsoft/code-push/issues/806

### DIFF
--- a/src/script/code-push-error.ts
+++ b/src/script/code-push-error.ts
@@ -1,13 +1,19 @@
+import {Http} from "./acquisition-sdk";
+
 export class CodePushError extends Error {
-    constructor(message: string) {
-        super(message);
+    constructor(message: string, options?: unknown) {
+        super(
+            message,
+            // @ts-expect-error - need to enable es2022 (see: https://github.com/microsoft/TypeScript/blob/main/src/lib/es2022.error.d.ts)
+            options
+        );
         Object.setPrototypeOf(this, CodePushError.prototype);
     }
 }
 
 export class CodePushHttpError extends CodePushError {
-    constructor(message: string) {
-        super(message);
+    constructor(message: string, response: Http.Response) {
+        super(message, {cause: response});
         Object.setPrototypeOf(this, CodePushHttpError.prototype);
     }
 }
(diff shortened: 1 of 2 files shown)


retyui:
@microsoft-github-policy-service agree

src/script/code-push-error.ts
```diff
@@ -6,8 +8,12 @@ export class CodePushError extends Error {
 }
 
 export class CodePushHttpError extends CodePushError {
-    constructor(message: string) {
-        super(message);
+    constructor(message: string, response: Http.Response) {
+        super(
+            message,
+            // @ts-expect-error - need to enable es2022 (see: https://github.com/microsoft/TypeScript/blob/main/src/lib/es2022.error.d.ts)
+            {cause: response}
```
DmitriyKirakosyan, line 15: How does it work? There is no second parameter in `CodePushError` class.

retyui: you right, I fixed

src/script/code-push-error.ts
```diff
@@ -1,13 +1,19 @@
+import {Http} from "./acquisition-sdk";
+
 export class CodePushError extends Error {
-    constructor(message: string) {
-        super(message);
+    constructor(message: string, options?: unknown) {
+        super(
+            message,
+            // @ts-expect-error - need to enable es2022 (see: https://github.com/microsoft/TypeScript/blob/main/src/lib/es2022.error.d.ts)
```
DmitriyKirakosyan, line 7: What would happen if we use es prior es2022? Should we add a fallback?

retyui: @DmitriyKirakosyan hi

velimir-jankovic: @retyui, let's keep compatibility with older ES.

My suggestion is to add a new member to `CodePushError` class, and then set this member in constructor, rather than `option` parameter in `Error` class. Same thing in `CodePushHttpError` derived class.

This way we encapsulate error in our own class rather than depend on new features of ES2022.

PR #819, opened 2023-04-18
Version 4.1.1

AnatolyPristensky:


denysop (approved):
LGTM

plugin.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<plugin xmlns="http://apache.org/cordova/ns/plugins/1.0" id="code-push" version="4.0.5">
+<plugin xmlns="http://apache.org/cordova/ns/plugins/1.0" id="code-push" version="4.1.1">
```
lucen-ms, line 2: Why do we change cordova plugin versin?

AnatolyPristensky: it's just for consistency.

Issue #811, opened 2023-01-17
4.1.0 has no d.ts files


pjc0247:
https://www.npmjs.com/package/code-push?activeTab=explore

4.0.5 works fine.

pjc0247:
a

saurav0705:
any update on this?

saurav0705:
Please find the pr for the above issue https://github.com/microsoft/code-push/pull/827

regalstreak:
bumping ^ @DmitriyKirakosyan

DmitriyKirakosyan:
@regalstreak , @saurav0705 , @pjc0247 ,

I tried to find the problem with d.ts files not being created using Node 12 and 16, but everything worked fine for me.

Can you tell me exactly what steps you took? This will help me understand the issue better.

Also, I looked at versions 4.0.5 and 4.1.0 and didn't see any big changes that could cause this problem. If you noticed something, please share.

regalstreak:
@DmitriyKirakosyan try creating a new typescript project and adding the latest version of code-push, the package misses the declarations (checked in node_modules)

4.0.5 also didn't have the d.ts for me

saurav0705:
@DmitriyKirakosyan the issue the release is pushed with the `npm run build` command due to which no `.d.ts` files are generated and package is published with that command

velimir-jankovic:
@saurav0705 thank you for the help! We've fixed our internal release process, so next release will include definition files properly.

victor-chan-groundswell:
I'm actually running into this right now, going from 4.0.5 to 4.2.0..... I double checked what's in the node_modules directory and the .d.ts files are definitely missing in the new version.

PR #827, opened 2023-07-12
fix: added ci check for declaartions and added fix for the same

saurav0705:
## [Issue](https://github.com/microsoft/code-push/issues/811)
In the latest version for [code-push web sdk](https://www.npmjs.com/package/code-push?activeTab=code) no declarations files are committed due to which ts is failing.

## Fix
Added `outDir` addition option in `tsconfig-release.json` which fixes the issue
image

## Prevention
Added a build check so that this should not happen in future releases

image

### DIFF
--- a/.github/scripts/check-for-declaration-files.ts
+++ b/.github/scripts/check-for-declaration-files.ts
@@ -1,0 +1,44 @@
+import fs from "fs";
+import path from "path";
+
+type ResultType = {
+    js : Record<string , boolean>
+    ts : Record<string , boolean>
+}
+
+const result: ResultType = {js:{} , ts:{}}
+
+const readThroughDirectory = (directory: string): void => {
+  const __directoryPath = path.join(process.env.INIT_CWD ?? '', directory);
+  const files = fs.readdirSync(__directoryPath);
+  files.forEach((file) => {
+    const filePath = path.join(__directoryPath, file);
+    const stats = fs.statSync(filePath);
+    if (stats.isDirectory()) {
+      readThroughDirectory(filePath);
+    }
+
+    if(filePath.endsWith('.js')){
+        const name = filePath.split('.')
+        name.pop()
+        result.js[name.join('.')] = true
+    }
+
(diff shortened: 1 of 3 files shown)


DmitriyKirakosyan:
@saurav0705 , the check passes for me even without the change (adding outDir) in this PR.

saurav0705:
> @saurav0705 , the check passes for me even without the change (adding outDir) in this PR.

Hi @DmitriyKirakosyan, yes I checked this I was able to reproduce this previously and not now so might be a cache issue but can we add this check so this doesn't happen in future

DmitriyKirakosyan:
@saurav0705 , if the check doesn't test the added changes, it might be more confusing than helpful. It would be great if you could modify the check to ensure it fails when `outDir` is absent.

saurav0705:
> @saurav0705 , if the check doesn't test the added changes, it might be more confusing than helpful. It would be great if you could modify the check to ensure it fails when `outDir` is absent.

@DmitriyKirakosyan issue is that the package that has been published to `npmjs` has been generated using command `npm run build` and not `npm run build:release` due to which compiler options don't have the `declarations : true` due to which the `.d.ts` file is not generated

I think making an action that automatically publishes to `npmjs` so that this doesn't happen

velimir-jankovic:
> > @saurav0705 , if the check doesn't test the added changes, it might be more confusing than helpful. It would be great if you could modify the check to ensure it fails when `outDir` is absent.
> 
> @DmitriyKirakosyan issue is that the package that has been published to `npmjs` has been generated using command `npm run build` and not `npm run build:release` due to which compiler options don't have the `declarations : true` due to which the `.d.ts` file is not generated
> 
> I think making an action that automatically publishes to `npmjs` so that this doesn't happen Also can you look into this #828 as this is the latest PR and also have a command that can be run before publishing to artifactory

As already discussed, changes in this PR won't ensure release without definition files will be created. 
We have internal pipeline for releases, I will make necessary changes there.

Closing.

PR #851, opened 2024-11-14
Disable api calls

DordeDimitrijev:
## Description
Implemented feature for stopping api calls from being sent after unsuccessful response. This behavior is aligned with other appcenter SDKs. Disabling of api calls occurs only per session after first failed request.

## Related PRs or issues
[AB#107980](https://dev.azure.com/msmobilecenter/Mobile-Center/_workitems/edit/107980)

PR #852, opened 2024-12-16
Bump version to 4.2.3

DordeDimitrijev:
Changed included in this release:

- https://github.com/microsoft/code-push/pull/851

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "code-push",
-  "version": "4.2.2",
+  "version": "4.2.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "code-push",
-      "version": "4.2.2",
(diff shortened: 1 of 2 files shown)


MikhailSuendukov:
Maybe we should add an article about API disabling behavior to README?

lucen-ms (approved):

lucen-ms:
> Maybe we should add an article about API disabling behavior to README?

do we describe it in other SDKs?

DordeDimitrijev:
> > Maybe we should add an article about API disabling behavior to README?
> 
> do we describe it in other SDKs?

I don't see it being mentioned in other SDKs. I don't feel like we need to explicitly mention it in README since we aligned it with existing behavior from other SDKs?

DmitriyKirakosyan (approved):

MikhailSuendukov (approved):

